PR #91, opened 2022-01-13
8268231: Aarch64: Use Ldp in intrinsics for String.compareTo

dchuyko:
Original patch applies cleanly.

Benchmark results of test/micro/org/openjdk/bench/java/lang/StringCompareToDifferentLength.java correspond to original issue, 10-25% 17u improvements on Graviton 2 for long enough strings (72+ chars).

Testing: tier1, tier2 (aarch64).

---------
### Progress

### Issue
 * [JDK-8268231](https://bugs.openjdk.java.net/browse/JDK-8268231): Aarch64: Use Ldp in intrinsics for String.compareTo

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17u-dev pull/91/head:pull/91` \
`$ git checkout pull/91`

`$ git checkout pull/91` \
`$ git pull https://git.openjdk.java.net/jdk17u-dev pull/91/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 91`

`$ git pr show -t 91`

</details>

<a href="https://git.openjdk.java.net/jdk17u-dev/pull/91.diff">https://git.openjdk.java.net/jdk17u-dev/pull/91.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
@@ -4915,18 +4915,6 @@ class StubGenerator: public StubCodeGenerator {
     return entry;
   }
 
-  // code for comparing 16 bytes of strings with same encoding
-  void compare_string_16_bytes_same(Label &DIFF1, Label &DIFF2) {
-    Register result = r0, str1 = r1, cnt1 = r2, str2 = r3, tmp1 = r10, tmp2 = r11;
-    __ ldr(rscratch1, Address(__ post(str1, 8)));
-    __ eor(rscratch2, tmp1, tmp2);
-    __ ldr(cnt1, Address(__ post(str2, 8)));
-    __ cbnz(rscratch2, DIFF1);
-    __ ldr(tmp1, Address(__ post(str1, 8)));
-    __ eor(rscratch2, rscratch1, cnt1);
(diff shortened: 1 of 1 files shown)


adinn:
I am not a maintainer so it is not in the end my call as to whether this goes in. However, I am against backporting this change.

I recall the original discussion where the conclusion was that this patch only managed to scrape is way into jdk18 because

1. it caused no visible harm
2. it improved a rare use case on some micro-architectures

That's maybe ok as a reason to commit to a dev branch but to me it's a poor reason to backport.

Standing against whatever limited gain this might offer are not just the risk of destabilising existing deployments that is present with any unnecessary change but also an increase in code size for the intrinsic. That latter outcome may not show in the micro-benchmark results but it may well increase icache pressure, which in itself might undermine the validity of point 1 above.

dchuyko:
I think it is actually a win for many actual aarch64 implementations. For example here is a chart for L-L string comparison on A72 (RPi4) up to length 128. Starting from size 72 there is a constant 16-17% time reduction. If we go to really long strings there which can be considred a rare use, the improvement grows to 38% for size 256 and even over 50% for size 512.
image

adinn:
@dchuyko @phohensee Has this patch been approved by one of the maintainers?

adinn:
Ok, I see Goetz commented on the JIRA issue.

PR #469, opened 2022-06-15
Backport cd8709e8e05897d131afba221970c0866b3d126d

GoeLin:
I backport this for parity with 17.0.5-oracle.

I had to resolve because "8275047: Optimize existing fill stubs for AVX-512 target"
which renames fill**_avx() to fill**() is not in 17.


TheRealMDoerr (approved):
Ok, the new `fill` functions are only used inside MacroAssembler. LGTM.

PR #635, opened 2022-08-17
8256811: Delayed/missed jdwp class unloading events

adinn:
Backport of patch to fix class unloading event notification.

This backport was originally developed by @zhengyu123 and involved some minor changes to the original patch.

n.b. the original patch required a follow up fix (cf JDK-8290908) which also needs backporting after this has been included.

---------
### Progress

### Issue
 * [JDK-8256811](https://bugs.openjdk.org/browse/JDK-8256811): Delayed/missed jdwp class unloading events

### Reviewers
 * [Andrew John Hughes](https://openjdk.org/census#andrew) (@gnu-andrew - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev pull/635/head:pull/635` \
`$ git checkout pull/635`

`$ git checkout pull/635` \
`$ git pull https://git.openjdk.org/jdk17u-dev pull/635/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 635`

`$ git pr show -t 635`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/635.diff">https://git.openjdk.org/jdk17u-dev/pull/635.diff</a>

</details>


RealCLanger:
@adinn: could you please enable GitHub actions on your repository and trigger a run via the workflow-dispatch event? Same goes for #636. Thanks!

zhengyu123:
This fix results a couple of followup CRs: JDK-8290908 and JDK-8291456. JDK-8291456 has yet been resolved, probably this backport should wait till JDK-8291456 is resolved.

adinn:
@RealCLanger The settings for my repo look like they already enable actions. Under Actions/General/Actions Permissions the radio button is selected for Allow all actions and reusable workflows. Do I need to set something else?

What do you mean by 'trigger a run via the workflow-dispatch event'?

adinn:
@zhengyu123 Thanks for the heads-up regarding JDK-8291456. I was not aware there was another dependent issue.

adinn:
@RealCLanger Ah, ok, I think I see now what you mean. I merged master and pushed and I can see that workflows are now running for the updated branch.

gnu-andrew (commented):
Backport mostly looks clean. The changes in `JvmtiExport::post_object_free` are appropriate, given the additions in JDK 19+ for virtual threads that are not present in 17.

I'm not 100% on the copyright additions which are not part of the original patch, as they may interfere with other backports. However, I guess they are needed at some point.

Most curious to me is this addition:

~~~
+-    JvmtiThreadState *state = thread->as_Java_thread()->jvmti_thread_state();
++    JvmtiThreadState *state = ((JavaThread*)thread)->jvmti_thread_state();
~~~

Is there a reason for altering this line? It doesn't seem related to any of the other changes, and, while the direct cast may be more performant than the method call, the existing usage with the method call seems safer.

adinn:
> Most curious to me is this addition:
> 
> +-    JvmtiThreadState *state = thread->as_Java_thread()->jvmti_thread_state();
> ++    JvmtiThreadState *state = ((JavaThread*)thread)->jvmti_thread_state();
> 
> Is there a reason for altering this line? It doesn't seem related to any of the other changes, and, while the direct cast may be more performant than the method call, the existing usage with the method call seems safer.

No, there is no reason for it other that that it was in the patch Zhengyu originally started working on. Of course, the cast is safe because of the guard in the preceding `if` test. However, that's not a reason to put it into the backport. I'm not it even helps performance since as_Java_thread() just does a static cast. I'll remove it.

gnu-andrew (approved):
Thanks, the rest looks good.

Note that [JDK-8291456](https://bugs.openjdk.org/browse/JDK-8291456) seems to mainly be about the test and has spawned two related test fixes, [JDK-8292880](https://bugs.openjdk.org/browse/JDK-8292880) and [JDK-8291650](https://bugs.openjdk.org/browse/JDK-8291650), which may be worth a look.

gnu-andrew:
Next step is to label the bug with `jdk17u-fix-request` and wait for `jdk17u-fix-yes` to be added by a maintainer.

adinn:
Thanks for the review @gnu-andrew

> Note that [JDK-8291456](https://bugs.openjdk.org/browse/JDK-8291456) seems to mainly be about the test and has spawned two related test fixes, [JDK-8292880](https://bugs.openjdk.org/browse/JDK-8292880) and [JDK-8291650](https://bugs.openjdk.org/browse/JDK-8291650), which may be worth a look.

Yes, I am aware of both of those follow-up issues. Although they both try to fix the test and, indeed, both appear to have lowered the chance of a failure, neither of them was 100% successful. So, I have been waiting for a proper test fix upstream (which is pending but seems not to be imminent). I was also hoping that I could maybe help out with an upstream fix but have not had the time to do so.

We could continue on that course or we could go ahead and push both these patches as is and ignore any failure in the test. Do you have any opinion on how to proceed?

gnu-andrew:
My impression from the bugs was that the code change itself is an improvement, and fewer events are missed, but it is difficult to get a reliable test which triggers class unloading and picks up all the events. I'm tending towards going ahead with backporting the four existing patches, but I expect you have a better idea of the risk than I do.

If we do this now, it'll be targeted for the January release so we still have some time to add further follow-up fixes or even revert in 17u. I don't get the impression that the main patch is flawed enough to warrant being reverted in trunk, but if we want to be extra safe, we could wait until some time after OpenJDK 20 is released next March.

adinn:
The upstream work on this issue is still in progress as fixes up to now have improved but neither eradicated the problem nor even diagnosed root cause - in fact the fixes actually stumbled into a secondary issue on Windows. So, I have been hesitating before pushing something that might not remove or workaround the problem.

It looks like Chris Plummer has identified the source of the Windows issue and provided a workaround for it (see https://bugs.openjdk.org/browse/JDK-8292879 and associated PR https://github.com/openjdk/jdk/pull/10519). So, I think it is probably worth proceeding with all the fixes up to and including this one.

adinn:
Correction: JDK-8291456 needs to be added after JDK-8291650 as it removes the delay added by JDK-8291650.

adinn:
Aargh! It seems JDK-8291456 was applied last after the other 3 commits. So, the required order is:

 - _JDK-8256811_ : Zhengyu's original fix
 - _JDK-8290908_ : Correct TCK failure from last patch
 - _JDK-8291650_ : Add delay to ClassUnloadEventTest to allow ObjectFree events to propagate
 - _JDK-8292880_ : Enable debuggee logging in ClassUnloadEventTest
 - _JDK-8292879_ : Add delay to skew race between ClassPrepare event propagation and GC request initiation in debuggee
 - _JDK-8291456_ : Flush pending ObjectFree events at VM shutdown

GoeLin:
@adinn, It seems they all miss the Fix request comment in the JBS issue. 
While the other ones are trivial, for this one please state reason and risk.  Thanks.

adinn:
Fix Request

The failure to correctly report class unloading events both in the OpenJDK tests and in actual deployments is causing spurious failures for the eclipse team's testing regime. This initial patch and the follow up patches mentioned above were adopted as remedies in upstream releases. Although they are still only an ad hoc fix, do in practice avoid the OpenJDK test failures and should produce the same results for other users.

Some of the upstream patches were superseded by later ones in order to arrive at a satisfactory solution. However, rather than create a new patch to arrive at the same end result it seems best to backport them all in sequence. They all apply cleanly apart from the first one.

One patch merely enables debug output in a relevant test. That has been retained, 1) to maintain patch compatibility with upstream, 2) to ensure follow-on patches apply cleanly, 3) because it will be useful if the ad hoc nature of the current fixes means further problems are encountered and 4) because its effect is minimal, being local only to the test and only resulting in a small amount more info in the test output file.

The overall risk is low because these patches only modify the class unloading code.

The changes have been tested by running the relevant class unloading unit tests. This does not guarantee the problem is fixed (because it is intermittent). However, it does indicate that the patches are not flawed.

adinn:
@GoeLin Apologies for forgetting to include Fix request comments. See above for a comment that covers the full set of patches.

gnu-andrew:
Thanks both. I've approved 8256811. I see @GoeLin already approved the others, so this should all be good to go.

As they are clean, it should be possible to  integrate the others once testing on the PR completes successfully.

adinn:
> Thanks, the reasoning is fine, but it usually goes into the JBS issue.
> I copied it there.

@GoeLin Apologies for my all too obviously rusty knowledge of the backport process and many thanks for fixing up my mistakes. Clearly, I have been spending too much time working on the GraalVM project.

@gnu-andrew Thanks for checking all the PRs. The test failures on this PR don't seem to be anything to do with the changes in the patch. The Windows one appears to be a (transient?) download failure during the gate run. The Mac failures are all timeouts (in TestUnsignedByteCompare1, TestTrichotomyExpressions and TestLinearScanOrderMain, none of which will can anything to do with this change). Is another run needed? If so is there a simple way to force that to happen?

gnu-andrew:
I agree with your analysis and I don't think another run is necessary. I was thinking more of the follow-on PRs, as, being clean, they don't need to wait on a review. I now see you've already filed these, and, given that the tests on those pass (https://github.com/adinn/jdk17u-dev/actions/) with this change included, that seems to be enough.

For reference, you should be able to re-run the job from the button on the right-hand side of https://github.com/adinn/jdk17u-dev/actions/runs/3006655779

adinn:
@gnu-andrew @GoeLin Ok, so does that mean I now proceed to integrate this PR and the dependent ones?

GoeLin:
Yes, feel free!
I had also checked the tests, to me they look unrelated, too.
You must give the bots some time to work on the dependent PRs. 
They will change the branch to compare to to master etc.
Sometimes they ask for a trivial but annoying resolve.

PR #853, opened 2022-11-02
8289524: Add JFR JIT restart event

MBaesken:
backport of 8289524

---------
### Progress

### Issue
 * [JDK-8289524](https://bugs.openjdk.org/browse/JDK-8289524): Add JFR JIT restart event

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev pull/853/head:pull/853` \
`$ git checkout pull/853`

`$ git checkout pull/853` \
`$ git pull https://git.openjdk.org/jdk17u-dev pull/853/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 853`

`$ git pr show -t 853`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/853.diff">https://git.openjdk.org/jdk17u-dev/pull/853.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/code/codeCache.cpp
+++ b/src/hotspot/share/code/codeCache.cpp
@@ -1313,6 +1313,7 @@ void CodeCache::report_codemem_full(int code_blob_type, bool print) {
     event.set_adaptorCount(heap->adapter_count());
     event.set_unallocatedCapacity(heap->unallocated_capacity());
     event.set_fullCount(heap->full_count());
+    event.set_codeCacheMaxCapacity(CodeCache::max_capacity());
     event.commit();
   }
 }
--- a/src/hotspot/share/jfr/metadata/metadata.xml
+++ b/src/hotspot/share/jfr/metadata/metadata.xml
@@ -494,6 +494,11 @@
     <Field type="ulong" contentType="bytes" name="used" label="Used" />
   </Event>
 
+  <Event name="JitRestart" category="Java Virtual Machine, Compiler" label="JIT Restart" stackTrace="false" startTime="false" thread="true">
+    <Field type="int" contentType="bytes" name="freedMemory" label="Freed Memory" />
+    <Field type="ulong" contentType="bytes" name="codeCacheMaxCapacity" label="Code Cache Maximum Capacity" />
+  </Event>
+
   <Event name="Compilation" category="Java Virtual Machine, Compiler" label="Compilation" thread="true" commitState="_thread_in_native">
     <Field type="uint" name="compileId" label="Compilation Identifier" relation="CompileId" />
     <Field type="CompilerType" name="compiler" label="Compiler" />
@@ -548,6 +553,7 @@
     <Field type="int" name="adaptorCount" label="Adaptors" />
     <Field type="ulong" contentType="bytes" name="unallocatedCapacity" label="Unallocated" />
     <Field type="int" name="fullCount" label="Full Count" />
+    <Field type="ulong" contentType="bytes" name="codeCacheMaxCapacity" label="Code Cache Maximum Capacity" />
   </Event>
 
   <Event name="Deoptimization" category="Java Virtual Machine, Compiler" label="Deoptimization" thread="true" stackTrace="true" startTime="false">
--- a/src/hotspot/share/runtime/sweeper.cpp
+++ b/src/hotspot/share/runtime/sweeper.cpp
@@ -406,11 +406,6 @@ void NMethodSweeper::sweep_code_cache() {
     _peak_sweep_time = MAX2(_peak_sweep_time, _total_time_this_sweep);
   }
 
-  EventSweepCodeCache event(UNTIMED);
-  if (event.should_commit()) {
-    post_sweep_event(&event, sweep_start_counter, sweep_end_counter, (s4)_traversals, swept_count, flushed_count, zombified_count);
-  }
-
 #ifdef ASSERT
   if(PrintMethodFlushing) {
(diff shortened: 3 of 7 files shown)


egahlin:
This is an enhancement. The bar for backporting events should be high. This PR also changes metadata of an existing event (adds a new field) which increases the risk for issues. The event is enabled by default, so it will have an impact on existing recordings running in production. It also contains a bug (JDK-8295419, fixed in mainline).

MBaesken:
I am aware of the name change done in JDK-8295419, would backport this afterwards.
Regarding
>This PR also changes metadata of an existing event 
Yes sure it does , if it is such a bad concern it would be possible to separate the CodeCacheFull related change from the backport, not sure if this is really an issue in practise.

egahlin:
Why is this enhancement needed in an update release? 

I don't doubt the event has some use, all have or they would not be integrated into mainline. I just don't see what makes this event special. If a user upgrades to get the latest security fixes, and this event (or similar) is included and for some reason causes issues, how can it be explained? Why is it so important?

MBaesken:
Having a couple of additional interesting (and not too complex !) events in the latest LTS is not a bad thing but this is my personal impression.  On the other hand I do not care that much, if it is rejected than probably all additional events would be rejected because the others I looked at that were interesting were more complex.
Our users will avoid non-LTS releases because of the effort and investment costs , so this means Sept 2023 + probably ~ one more year of testing etc. , until they can consume other interesting JFR events that came in after 17 , quite a long time I think.

egahlin:
> Having a couple of additional interesting (and not too complex !) events in the latest LTS is not a bad thing but this is my personal impression. On the other hand I do not care that much, if it is rejected than probably all additional events would be rejected because the others I looked at that were interesting were more complex.

I think the bar should be high and there should be a good reason, for example to support new hardware or OS versions. Improve security.  Things that are essential to be able to use the existing JDK release.  I don't see this JFR event fall into that category.

RealCLanger:
> > Having a couple of additional interesting (and not too complex !) events in the latest LTS is not a bad thing but this is my personal impression. On the other hand I do not care that much, if it is rejected than probably all additional events would be rejected because the others I looked at that were interesting were more complex.
> 
> I think the bar should be high and there should be a good reason, for example to support new hardware or OS versions. Improve security. Things that are essential to be able to use the existing JDK release. I don't see this JFR event fall into that category.

I know there have been discussions about backporting JFR events earlier on and I think none have ever happened. But I think we should constantly reconsider our decisions. Regarding JFR event backports, we have been very conservative overall and I think that is not bad. We need to be careful regarding stability and not to break things. Also, major feature improvements need to come with newer JDK releases. However, we still need and want to support our customers that are running on LTS releases of the OpenJDK. And that's what additional/improved JFR events will help to achieve. So, while we should err on the path of caution as to not break anything, I think we should not reject JFR enhancements per se. At least that's my current thinking...

tstuefe:
@egahlin do you have any concrete concerns regarding the safety or risk of this patch? Or would it make the produced JFR output incompatible with JMC or other readers in any way?

If it is just a general concern about downporting practices, I vote for downporting it. As maintainers we have a high interest in keeping 17u stable, but we also need to have tools to analyze problems in the field.

MBaesken:
> I know there have been discussions about backporting JFR events earlier on and I think none have ever happened. 

We had a few backports to 11  ([8003209](https://bugs.openjdk.org/browse/JDK-8003209) even backported to 8) in the past

https://bugs.openjdk.org/browse/JDK-8213617
(11.0.4)

https://bugs.openjdk.org/browse/JDK-8223438
(11.0.6)

https://bugs.openjdk.org/browse/JDK-8003209
(11.0.1)

The last one looks like a rather complex event, as far as I can see ...
Two of my examples added events, one just a field to an existing event. Not aware of any tools crashing/stopped working because of those additions (but to be fair, I do not monitor all tools that deal with JFR events).
See also 
https://github.com/openjdk/jdk11u-dev/commits/master/src/hotspot/share/jfr/metadata/metadata.xml
to get more info about the topic.

Maybe regarding this event, as a compromise we could wait until jdk20 is released (includes JDK-8289524) and then some time;  and then do the backport.

egahlin:
> @egahlin do you have any concrete concerns regarding the safety or risk of this patch? Or would it make the produced JFR output incompatible with JMC or other readers in any way?
> 

I don't know the impact on JMC or other tools consuming JFR data.

To know, one would need to try them out, but any modification of the event metadata, or what is enabled in the default configuration, risks being a problem. 

When backporting events several releases back, it also becomes confusing for JFR API users since the event will be missing in release between. In this case, JDK 17 and JDK 18.

> If it is just a general concern about downporting practices, I vote for downporting it. As maintainers we have a high interest in keeping 17u stable, but we also need to have tools to analyze problems in the field.

I like to know what problem this event solves? 

If there have been say 3-5 JVM bugs where this event would have resolved them significant faster, I think an argument can be made (and explained to a customer why it was added in a security update). It contributes to stability (faster fixes). On the other hand, adding events to get features 10 months faster into a LTS release doesn't sound like the right reason to me.

TheRealMDoerr:
The `EventJitRestart` is the counterpart of the `EventCodeCacheFull`. Those who want to see the latter will probably also want to see the first one. Otherwise, you can't see when the problem is solved (performance issue or other issues related to full code cache). It may be needed to understand other events.
It is legitimate to omit non-LTS releases, but not so nice. JDK 18 is out of support.
I think a JFR user should try it and comment about the usefulness.

MBaesken:
it is labeled clean so I got the info from Goetz Lindenmaier that it can be integrated.

PR #878, opened 2022-11-10
Backport 61698bd137fc8ffad6a63b57b73df699712527b9

dchuyko:
Original patch does not apply because of different NOT_PRODUCT fields of InterpreterCodelet. Repeated exactly the meaningful part: moving 3 fields in a new order after the NOT_PRODUCT part. Comments and white space 'after' are just like in the original diff.

Testing: tier1, tier2 on aarch64.


phohensee (approved):
Lgtm.

PR #920, opened 2022-11-22
Backport 3f2f128af6ec2f9097af7758bfd41aeaa4354d40

zzambers:
Problemlist should be extended so that existing compiler/rtm entries include x86 (32-bit) intel builds as well, as these are also affected.


zzambers:
One build failure caused by failed checksum on asmtools sources (?) does not seem to have any relation to this change.

PR #931, opened 2022-11-25
Backport 470f3424fcce0e41b75cccdd5e3a56771cd07ff5

RealCLanger:


MBaesken (commented):
We always check TOOLCHAIN_PATH in basic.m4 
UTIL_LOOKUP_PROGS(XCODEBUILD, xcodebuild, $TOOLCHAIN_PATH)
Shouldn't we do it similar to jdk/jdk  ?  There we check in case of DEVKIT_ROOT set, DEKIT_TOOLCHAIN_PATH.  In case of  TOOLCHAIN_PATH set we check TOOLCHAIN_PATH.

RealCLanger:
> We always check TOOLCHAIN_PATH in basic.m4 UTIL_LOOKUP_PROGS(XCODEBUILD, xcodebuild, $TOOLCHAIN_PATH) Shouldn't we do it similar to jdk/jdk ? There we check in case of DEVKIT_ROOT set, DEKIT_TOOLCHAIN_PATH. In case of TOOLCHAIN_PATH set we check TOOLCHAIN_PATH.

Hm, maybe yes. But in that case we should have a look at backporting [JDK-8287254](https://bugs.openjdk.org/browse/JDK-8287254) entirely. For the current level of jdk17u, I would think this fits.

MBaesken:
Okay thanks for the explanation. Backporting [JDK-8287254](https://bugs.openjdk.org/browse/JDK-8287254)  is a bit too much for now (maybe later?).

MBaesken (approved):

PR #1048, opened 2023-01-10
Backport 3ff085e2967508ad312c9d32fa908807aefe69ee

tstuefe:
I'd like to downport this, since it is a prerequisite for downporting [JDK-8275704](https://bugs.openjdk.org/browse/JDK-8275704)
"Metaspace::contains() should be threadsafe".

### DIFF
--- a/src/hotspot/share/memory/metaspace/chunkManager.cpp
+++ b/src/hotspot/share/memory/metaspace/chunkManager.cpp
@@ -316,24 +316,11 @@ void ChunkManager::purge() {
 
   const size_t reserved_before = _vslist->reserved_words();
   const size_t committed_before = _vslist->committed_words();
-  int num_nodes_purged = 0;
 
-  // We purge to return unused memory to the Operating System. We do this in
-  //  two independent steps.
-
-  // 1) We purge the virtual space list: any memory mappings which are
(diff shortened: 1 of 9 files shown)


GoeLin:
Why is the Pre-submit test failing?

tstuefe:
> Why is the Pre-submit test failing? I remove the fix request tag in the meantime.

The 32-bit release build failed because of GHA infra problems. The 32-bit debug build succeeded though. But the x86 tests did not run because the release build was not finished (we may want to look into this: if debug built, debug should be tested even if release failed to build. Or do we need release for testing?)

Anyway, since the PR branch is owned by the Skara bot, I am not sure how to re-trigger the GHA tests. For my own branches, I just push a new empty commit. But how do I do this here? @RealCLanger do you know?

tstuefe:
I executed Metaspace related jtreg and -gtests on 32-bit debug locally, all is fine as expected. The patch was also tested in our nightlies for two nights in a row.

GoeLin:
Thanks, just to make sure it's not your change causing this.

GoeLin:
Theoretically, if you click on the Details link above and then go to the job, there is as redo button at the top right. 
But so far it never worked for me. I rather merge in head by navigating to the branch and clicking "sync-fork"

PR #1164, opened 2023-02-10
Backport a2f67660f088559ce49f73da7401801fb826028b

GoeLin:
I backport this for parity with 17.0.7-oracle.


GoeLin:
Pre-submit failure: A crash on x86!! Unrelated to this change.

Probably this should be investigated, but it is on 32-bit x86.

Error. Agent communication error: java.io.EOFException;
[libjvm.so+0x92ae4b]  Method::is_method_handle_intrinsic() const+0xb

PR #1211, opened 2023-03-21
Backport a00f5d24d3824e3ab84208401a967efe0e7bf88e

apavlyutkin:
Hi, here is backport of JDK-8303861. The patch fixes error reporting timeout processing for the case when a user specifies OnError  hook. Before the patch in this case JVM just ignored timeout and therefore WatcherThread could not break malloc() deadlock in error reporting routine.

If fact original patch applied without a scratch but there is a little conflics in the baselines, so the backport is not clean.

Verification/regression (amd64/LTS20.04):  test/hotspot/jtreg/runtime/ErrorHandling (includes updated test)

### DIFF
--- a/src/hotspot/share/runtime/nonJavaThread.cpp
+++ b/src/hotspot/share/runtime/nonJavaThread.cpp
@@ -270,8 +270,8 @@ void WatcherThread::run() {
           os::die();
         }
 
-        // Wait a second, then recheck for timeout.
-        os::naked_short_sleep(999);
+        // Wait a bit, then recheck for timeout.
+        os::naked_short_sleep(250);
       }
     }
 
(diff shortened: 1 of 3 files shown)


phohensee (approved):
Ltgm.

tstuefe (approved):
LGTM

phohensee:
JBS issue was tagged jdk17-fix-request, fixed.

GoeLin:
Hi,
in the test, you include two changes to Copyright that are not in the original change.
Why so?
Best, Goetz.

apavlyutkin:
My bad, original changeset includes only Red Hat record, dunno why I also copied SAP one that is out of the scope

PR #1330, opened 2023-05-08
8305088: SIGSEGV in Method::is_method_handle_intrinsic

TheRealMDoerr:
I had forgotten a null check when backporting [JDK-8295724](https://bugs.openjdk.org/browse/JDK-8295724). Unlike head, JDK17u uses `set_method(NULL)` in `nmethod::make_unloaded()` (and `nmethod::make_not_entrant_or_zombie`). We shouldn't ask `can_be_allocated_in_NonNMethod_space()` in these cases. The nmethod can't be a method handle intrinsic.

### DIFF
--- a/src/hotspot/share/code/nmethod.cpp
+++ b/src/hotspot/share/code/nmethod.cpp
@@ -1831,7 +1831,8 @@ bool nmethod::is_unloading() {
   // oops in the CompiledMethod, by calling oops_do on it.
   state_unloading_cycle = current_cycle;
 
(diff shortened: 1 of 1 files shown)


TheRealMDoerr:
Pre-submit error seems to be unrelated: sh][ERROR] wget exited with exit code 4

TheRealMDoerr:
Thanks for the review!
@TobiHartmann: You may also want to take a look. Seems like 17.0.7-oracle has the same issue.

TobiHartmann:
Thanks Martin! Yes, we will fix this in Oracle JDK as well.

PR #1414, opened 2023-06-01
Backport 438c969b7b07eeef0158b089e5a168849e04bf56

olivergillespie:
Hi all,

This pull request contains a backport of commit [438c969b](https://github.com/openjdk/jdk/commit/438c969b7b07eeef0158b089e5a168849e04bf56) from the [openjdk/jdk](https://git.openjdk.org/jdk) repository.

The commit being backported was authored by Sergey Tsypanov on 29 Mar 2023 and was reviewed by Naoto Sato.

Since the original uses the new `newHashMap` and `newHashSet` methods from https://bugs.openjdk.org/browse/JDK-8284780 which are not here in 17, I've not included those changes in the backport. They are not a significant part of the original fix, the main point of the change is kept which is avoiding the creation of 2 new hash sets on the cached path. Running the included benchmark confirms the performance improvement even with the omissions:

Base:
```
Benchmark                                                  Mode  Cnt      Score     Error   Units
ZonedDateTimeFormatterBenchmark.parse                      avgt   20  18248.082 ? 166.440   ns/op
ZonedDateTimeFormatterBenchmark.parse:?gc.alloc.rate       avgt   20   1274.964 ?  11.448  MB/sec
ZonedDateTimeFormatterBenchmark.parse:?gc.alloc.rate.norm  avgt   20  24400.418 ?   1.591    B/op
ZonedDateTimeFormatterBenchmark.parse:?gc.count            avgt   20     44.000            counts
ZonedDateTimeFormatterBenchmark.parse:?gc.time             avgt   20     35.000                ms
```

Patch:
```
Benchmark                                                  Mode  Cnt    Score   Error   Units
ZonedDateTimeFormatterBenchmark.parse                      avgt   20  923.943 ? 8.737   ns/op
ZonedDateTimeFormatterBenchmark.parse:?gc.alloc.rate       avgt   20  883.450 ? 8.163  MB/sec
ZonedDateTimeFormatterBenchmark.parse:?gc.alloc.rate.norm  avgt   20  856.021 ? 0.079    B/op
ZonedDateTimeFormatterBenchmark.parse:?gc.count            avgt   20   32.000          counts
ZonedDateTimeFormatterBenchmark.parse:?gc.time             avgt   20   27.000              ms
```

### DIFF
--- a/src/java.base/share/classes/java/time/format/DateTimeFormatterBuilder.java
+++ b/src/java.base/share/classes/java/time/format/DateTimeFormatterBuilder.java
@@ -4182,9 +4182,8 @@ public final class DateTimeFormatterBuilder {
             }
             Locale locale = context.getLocale();
             boolean isCaseSensitive = context.isCaseSensitive();
-            Set<String> regionIds = new HashSet<>(ZoneRulesProvider.getAvailableZoneIds());
-            Set<String> nonRegionIds = new HashSet<>(64);
-            int regionIdsSize = regionIds.size();
+            Set<String> availableZoneIds = ZoneRulesProvider.getAvailableZoneIds();
+            int regionIdsSize = availableZoneIds.size();
 
             Map<Locale, Entry<Integer, SoftReference<PrefixTree>>> cached =
                 isCaseSensitive ? cachedTree : cachedTreeCI;
@@ -4197,6 +4196,8 @@ public final class DateTimeFormatterBuilder {
                 (tree = entry.getValue().get()) == null)) {
                 tree = PrefixTree.newTree(context);
                 zoneStrings = TimeZoneNameUtility.getZoneStrings(locale);
+                Set<String> nonRegionIds = new HashSet<>(64);
+                Set<String> regionIds = new HashSet<>(availableZoneIds);
                 for (String[] names : zoneStrings) {
                     String zid = names[0];
                     if (!regionIds.remove(zid)) {
--- a/test/micro/org/openjdk/bench/java/time/format/ZonedDateTimeFormatterBenchmark.java
+++ b/test/micro/org/openjdk/bench/java/time/format/ZonedDateTimeFormatterBenchmark.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.openjdk.bench.java.time.format;
+
+import org.openjdk.jmh.annotations.*;
+
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.util.concurrent.TimeUnit;
+
+@Fork(4)
+@State(Scope.Thread)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Warmup(iterations = 5, time = 1)
+@Measurement(iterations = 5, time = 1)
+public class ZonedDateTimeFormatterBenchmark {
+
+    private static final DateTimeFormatter df = new DateTimeFormatterBuilder()
+            .appendPattern("yyyy:MM:dd:HH:mm:v")
+            .toFormatter();
+    private static final String TEXT = "2015:03:10:12:13:ECT";
+
+    @Setup
+    public void setUp() {
+        ZonedDateTime.parse(TEXT, df);
+    }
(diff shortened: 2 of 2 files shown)


shipilev:
You can still change `HashMap.newHashMap(1)` to `new HashMap(1)`, or am I missing something here?

olivergillespie:
> You can still change `HashMap.newHashMap(1)` in original backport to `new HashMap(1)`, instead of dropping them. Or am I missing something here?

Yes that's true, but it doesn't seem particularly valuable, and discussion on the details might easily outweigh any benefits. There wasn't much evidence provided for the original change, anyway.

For example, it should be `2` to match upstream behaviour due to the load factor. Using `1` appears to be slower (though lower alloc) than the default.

```
 @Benchmark
 public HashMap init0() {
     HashMap<String, String> x = new HashMap<>();
     x.put("1", "2");
     return x;
 }
 @Benchmark
 public HashMap init1() {
     HashMap<String, String> x = new HashMap<>(1);
     x.put("1", "2");
     return x;
 }
 @Benchmark
 public HashMap init2() {
     HashMap<String, String> x = new HashMap<>(2);
     x.put("1", "2");
     return x;
 }
```

```
Benchmark                              Mode  Cnt     Score     Error   Units
HashMapInit.init0                      avgt   20    24.655 ?   0.163   ns/op
HashMapInit.init0:?gc.alloc.rate       avgt   20  6187.811 ?  40.647  MB/sec
HashMapInit.init0:?gc.alloc.rate.norm  avgt   20   160.002 ?   0.006    B/op
HashMapInit.init0:?gc.count            avgt   20   136.000            counts
HashMapInit.init0:?gc.time             avgt   20   101.000                ms
HashMapInit.init1                      avgt   20    33.716 ?   0.492   ns/op
HashMapInit.init1:?gc.alloc.rate       avgt   20  3620.473 ?  51.771  MB/sec
HashMapInit.init1:?gc.alloc.rate.norm  avgt   20   128.000 ?   0.001    B/op
HashMapInit.init1:?gc.count            avgt   20   123.000            counts
HashMapInit.init1:?gc.time             avgt   20    86.000                ms
HashMapInit.init2                      avgt   20    22.776 ?   0.877   ns/op
HashMapInit.init2:?gc.alloc.rate       avgt   20  4361.898 ? 171.144  MB/sec
HashMapInit.init2:?gc.alloc.rate.norm  avgt   20   104.001 ?   0.004    B/op
HashMapInit.init2:?gc.count            avgt   20   116.000            counts
HashMapInit.init2:?gc.time             avgt   20    84.000                ms
```

I don't mind changing it if it's preferred, but I think the benefit is questionable so I lean toward the simpler change.

shipilev:
> Yes that's true, but it doesn't seem particularly valuable, and discussion on the details might easily outweigh any benefits. There wasn't much evidence provided for the original change, anyway.
> 
> For example, it should be `2` to match upstream behaviour due to the load factor. 

Right. It looks like `new HashMap<>(2)` is good then?

For backports, I would rather match the behavior to other releases as much as possible. At very least this allows to be more bug-to-bug compatible with other releases. If we match, a later bug/improvement report against this change in 17u have more chances to apply to other releases, and vice versa.

olivergillespie:
Thanks, yes `2` seems best for performance if we take it as correct that this usually has only one value (I have nothing to back that up except the comment in the original change), and we're happy to rely on knowledge of the default load factor.

Do you have a preference for:

- Use 2 with no comment
- Use 2 with comment
- Use something like `1/0.75f + 1` which is semi self-documenting

Judging from https://bugs.openjdk.org/browse/JDK-8186958, load factor is rarely considered when setting initial capacity in existing code. Which to me makes sense, because it's a private detail of the HashMap implementation, hence the new methods which handle those internals safely.

(I still prefer not specifying, but I'm happy to add whatever you think is best)

shipilev:
>     * Use 2 with comment

Yes, this should do:

```
        // Cache per instance for now. These HashMaps normally contain a single
        // element, initialize them with initial capacity = 2 to avoid resizes due to load factor.
        private final Map<Locale, Entry<Integer, SoftReference<PrefixTree>>>
            cachedTree = new HashMap<>(2);
        private final Map<Locale, Entry<Integer, SoftReference<PrefixTree>>>
            cachedTreeCI = new HashMap<>(2);
```

GoeLin:
Please first get the review before labeling!

src/java.base/share/classes/java/time/format/DateTimeFormatterBuilder.java
```diff
@@ -4197,6 +4197,8 @@ protected PrefixTree getTree(DateTimeParseContext context) {
                 (tree = entry.getValue().get()) == null)) {
                 tree = PrefixTree.newTree(context);
                 zoneStrings = TimeZoneNameUtility.getZoneStrings(locale);
+                Set<String> nonRegionIds = new HashSet<>(64);
```
shipilev, line 4200: Does the same argument about load factor applies here? So it should be not `64`, but something more? Load factor `0.75` implies it should be at least `86`. (The size would be rounded to the next power of two internally, so `128` might as well do.)

olivergillespie: It's not clear why 64 was chosen here. Reading https://bugs.openjdk.org/browse/JDK-8235238 I don't see anything which particularly justifies 64. On my system:

```
Set<String> regionIds = ZoneRulesProvider.getAvailableZoneIds();
Set<String> nonRegionIds = new HashSet<>();
for (String[] z : TimeZoneNameUtility.getZoneStrings(Locale.getDefault());) {
    if (!regionIds.contains(z[0])) nonRegionIds.add(z[0]);
}
System.out.println(nonRegionIds.size());
```

gives 28, so it seems 64 already accounts for load factor nicely at least on my system.

shipilev: Right. Actually, scratch my comment, since the current version matches what mainline does.

shipilev (approved):
Looks fine to me. Since this is not a very clean backport, I'll request another review.

/reviewers 2

src/java.base/share/classes/java/time/format/DateTimeFormatterBuilder.java
olivergillespie: While I'm happy you're happy, that's not quite right - mainline uses `HashSet.newHashSet(64)` which is equivalent to `new HashSet(128)`, but I don't think that was a justified change in mainline.

stsypanov (approved):
LGTM!

phohensee (approved):
Lgtm.

shipilev:
Push approval is there.

/integrate

shipilev:
Err.

/sponsor

PR #1573, opened 2023-07-12
Backport 959a61fdd483c9523764b9ba0972f59ca06db0ee

GoeLin:
I backport this to work around an error of msys2

I ran into this testing the bump to jtreg 7.2


GoeLin:
I removed setting the Java version to 17.
Obviously 8309934 and 8310183 have been tries to fix this issue here, which then had been undone partially by this change.
But our current setup does not work with jdk17.  There is unsupported applet stuff somewhere in the dependencies.
Anyways, the intention of this change is to work around the msys problem, and the remaining change does just this.

GoeLin:
GHA tests had passed before I removed the copyright change.

MBaesken (approved):

PR #1623, opened 2023-07-25
8312974: Bump update version for OpenJDK: jdk-17.0.10

GoeLin:
The usual updates.


yan-too:
Which means, new changes made in July/August/September would go to January not October release? I don't see openjdk wiki, it is under maintenance, but --

gnu-andrew:
> Which means, new changes made in July/August/September would go to January not October release? I don't see openjdk wiki, it is under maintenance, but -- oops, I see: to be pushed after the tag. But in the JBS there is "17.0.10 starts in repository jdk11u-dev after Tuesday August 1st, 2023" -- should it be August 30?

It should be according to the wiki; "Tuesday, August 29 2023: Rampdown; last merge from jdk17u-dev to jdk17u"

This PR seems very premature. It startled me as well.

PR #1630, opened 2023-07-28
8292989: Avoid dynamic memory in AsyncLogWriter

navyxliu:
This patch is *NOT* a clean backport.  The logic part is exactly the same, but I have to adjust code in the following parts:
1. change from KVHashtable to ResourceHashtable because we want to use resourceArea instead of C_HEAP. 
2. rewrite lambda expression using Functor. 

Testing: 
jdk-test1 tests including gtest. 
Manually check there's no malloc/free in runtime.

---------
### Progress
- [ ] [JDK-8292989](https://bugs.openjdk.org/browse/JDK-8292989) needs maintainer approval

### Issue
 * [JDK-8292989](https://bugs.openjdk.org/browse/JDK-8292989): Avoid dynamic memory in AsyncLogWriter (**Enhancement** - P4)

### Reviewers
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**) ⚠️ Review applies to [4511c52f](https://git.openjdk.org/jdk17u-dev/pull/1630/files/4511c52feb800cf0dbbbecc5e22c35fc247b6451)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/1630/head:pull/1630` \
`$ git checkout pull/1630`

`$ git checkout pull/1630` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/1630/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1630`

`$ git pr show -t 1630`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/1630.diff">https://git.openjdk.org/jdk17u-dev/pull/1630.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/1630#issuecomment-1655264069)

### DIFF
--- a/src/hotspot/share/logging/logAsyncWriter.cpp
+++ b/src/hotspot/share/logging/logAsyncWriter.cpp
@@ -27,6 +27,7 @@
 #include "logging/logFileOutput.hpp"
 #include "logging/logFileStreamOutput.hpp"
 #include "logging/logHandle.hpp"
+#include "memory/resourceArea.hpp"
 #include "runtime/atomic.hpp"
 #include "runtime/os.inline.hpp"
 
@@ -42,28 +43,49 @@ class AsyncLogWriter::AsyncLogLocker : public StackObj {
   }
 };
 
-void AsyncLogWriter::enqueue_locked(const AsyncLogMessage& msg) {
-  if (_buffer.size() >= _buffer_max_size) {
+// LogDecorator::None applies to 'constant initialization' because of its constexpr constructor.
+const LogDecorations& AsyncLogWriter::None = LogDecorations(LogLevel::Warning, LogTagSetMapping<LogTag::__NO_TAG>::tagset(),
+                                      LogDecorators::None);
+
+bool AsyncLogWriter::Buffer::push_back(LogFileStreamOutput* output, const LogDecorations& decorations, const char* msg) {
+  const size_t sz = Message::calc_size(strlen(msg));
+  const bool is_token = output == nullptr;
+  // Always leave headroom for the flush token. Pushing a token must succeed.
+  const size_t headroom = (!is_token) ? Message::calc_size(0) : 0;
+
+  if (_pos + sz <= (_capacity - headroom)) {
+    new(_buf + _pos) Message(output, decorations, msg);
+    _pos += sz;
+    return true;
+  }
+
+  return false;
+}
+
+void AsyncLogWriter::Buffer::push_flush_token() {
+  bool result = push_back(nullptr, AsyncLogWriter::None, "");
+  assert(result, "fail to enqueue the flush token.");
+}
+
+void AsyncLogWriter::enqueue_locked(LogFileStreamOutput* output, const LogDecorations& decorations, const char* msg) {
+  // To save space and streamline execution, we just ignore null message.
+  // client should use "" instead.
+  assert(msg != nullptr, "enqueuing a null message!");
+
+  if (!_buffer->push_back(output, decorations, msg)) {
     bool p_created;
-    uint32_t* counter = _stats.add_if_absent(msg.output(), 0, &p_created);
+    uint32_t* counter = _stats.put_if_absent(output, 0, &p_created);
     *counter = *counter + 1;
-    // drop the enqueueing message.
-    os::free(msg.message());
     return;
   }
 
-  _buffer.push_back(msg);
   _data_available = true;
   _lock.notify();
 }
 
 void AsyncLogWriter::enqueue(LogFileStreamOutput& output, const LogDecorations& decorations, const char* msg) {
-  AsyncLogMessage m(&output, decorations, os::strdup(msg));
-
-  { // critical area
-    AsyncLogLocker locker;
-    enqueue_locked(m);
-  }
+  AsyncLogLocker locker;
+  enqueue_locked(&output, decorations, msg);
 }
 
 // LogMessageBuffer consists of a multiple-part/multiple-line messsage.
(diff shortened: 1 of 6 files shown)


phohensee (approved):

phohensee:
windows-aarch64 build failure running adlc appears unrelated.

src/hotspot/share/logging/logAsyncWriter.hpp
```diff
@@ -135,7 +55,95 @@ typedef KVHashtable<LogFileStreamOutput*, uint32_t, mtLogging> AsyncLogMap;
 // change the logging configuration via jcmd, LogConfiguration::configure_output() calls flush() under the protection of the
 // ConfigurationLock. In addition flush() is called during JVM termination, via LogConfiguration::finalize.
 class AsyncLogWriter : public NonJavaThread {
+  friend class AsyncLogTest;
+  friend class AsyncLogTest_logBuffer_vm_Test;
   class AsyncLogLocker;
+  using AsyncLogMap = ResourceHashtable<LogFileStreamOutput*, uint32_t, primitive_hash<LogFileStreamOutput*>,
+                                        primitive_equals<LogFileStreamOutput*>, 17,
+                                        ResourceObj::C_HEAP, mtLogging>;
```
shipilev, line 63: These arguments should be on the new lines each, I think? This would match the upstream code better.

src/hotspot/share/utilities/resourceHash.hpp
```diff
@@ -184,11 +184,20 @@ class ResourceHashtable : public ResourceObj {
   // the iteration is cancelled.
   template<class ITER>
   void iterate(ITER* iter) const {
+    auto function = [&] (K& k, V& v) {
```
shipilev, line 187: Can we avoid doing this by making a named `struct` implementing `do_entry` for every use? I realize that would deviate the _uses_ from the upstream, but we would not be changing the shared code, which would be safer for generic code that does not use any of async logging *and* uses older toolchains.

This hunk effectively backports https://bugs.openjdk.org/browse/JDK-8276789, which seems to imply the need to check this works well with different toolchains.

shipilev (commented):
Looked through the code, I have comments.

navyxliu:
> Before I review the change, should https://bugs.openjdk.org/browse/JDK-8294075 be backported as well? Is JDK-8294075 caused by this change?

we found that it doesn't work. we still have intermittent failure. 
https://bugs.openjdk.org/browse/JDK-8309067

src/hotspot/share/utilities/resourceHash.hpp
navyxliu: yes, I can rewrite lambdas using old-style functor. 

to backport [JDK-8276789](https://bugs.openjdk.org/browse/JDK-8276789), we also need to backport bucket_at(), table()/table_size() beforehand. I take your advice. my approach skips more verification and it's not good.  I will use functor.

src/hotspot/share/logging/logAsyncWriter.hpp
navyxliu: hi, @shipilev 
Do we have linter rules for this?

navyxliu: even in tip of jdk, I still can find code like this: 
https://github.com/openjdk/jdk/blob/master/src/hotspot/share/classfile/classLoaderStats.hpp#L114

tstuefe:
Is it necessary to downport this patch? On its own, it does not sound that important, especially if we downport the glibc autotrimmer.

navyxliu:
Hi, Thomas, 

Thanks you for the comments. I was told that it is better reducing memory footprint of UL in jdk17u, a LTS Java. May I learn more about glibc autotrim? Is it a glibc feature or a new feature of hotspot? 

I found JDK-8268893 on JBS. It itself is a very useful feature! It is not an automatic trimmer, is it?

tstuefe:
Hi Xin :)

> Hi, Thomas,
> 
> Thanks you for the comments. I was told that it is better reducing memory footprint of UL in jdk17u, a LTS Java.

Interesting, I'm curious about the effectiveness of your patch. Do you see good results? I would have thought that a lot of these interleaving memory requests are just taken from and handed back into fastbins or tcache, and therefore don't make the arena grow. Also, the glibc attempts to reclaim memory automatically are quite limited (see PR description https://github.com/openjdk/jdk/pull/14781#issue-1790932923)

> May I learn more about glibc autotrim? Is it a glibc feature or a new feature of hotspot?
> 
> I found [JDK-8268893](https://bugs.openjdk.org/browse/JDK-8268893) on JBS. It itself is a very useful feature! It is not an automatic trimmer, is it?

Its https://bugs.openjdk.org/browse/JDK-8293114. You have to manually enable it, it is still marked as experimental. Also not in 17 yet, but Aleksey is preparing things to get it down quickly once it has baked in head a bit.

Essentially, what the trimmer does is run periodically a malloc_trim() operation. malloc_trim() will reclaim memory that has been free'd and that is in page granularity (so, it will also reclaim memory in the middle of an arena).

Cheers, Thomas

shipilev:
> Is it necessary to downport this patch? On its own, it does not sound that important, especially if we downport the glibc autotrimmer.

I think that autotrimmer is a mitigation strategy for oversubscription, rather than a fix. it also does not resolve fragmentation issues, if there are surviving mallocs between temporarly async logging mallocs. In other words, it is more compelling to avoid the issue instead of relying on auto-trimmer. We use a lot of `-Xlog:async` in production, and this would give us additional safety against memory leaks at least with glibc.

navyxliu:
hi, @tstuefe, 

I think we need both of them. This patch refrains from using any malloc/free in the first place. It also helps avoid the fragmentation issue.  Autotrimmer is a general solution for glibc overprovision.  

Should we still consider to include this patch in jdk17u?

tstuefe:
> hi, @tstuefe,
> 
> I think we need both of them. This patch refrains from using any malloc/free in the first place. It also helps avoid the fragmentation issue. Autotrimmer is a general solution for glibc overprovision.
> 
> Should we still consider to include this patch in jdk17u?

I don't want to block this patch, so sure. I just would have liked to know what the strategy brings in terms of RSS. Plain curiousity.

src/hotspot/share/logging/logAsyncWriter.cpp
```diff
@@ -42,28 +43,49 @@ class AsyncLogWriter::AsyncLogLocker : public StackObj {
...
+
+void AsyncLogWriter::enqueue_locked(LogFileStreamOutput* output, const LogDecorations& decorations, const char* msg) {
+  // To save space and streamline execution, we just ignore null message.
+  // client should use "" instead.
+  assert(msg != nullptr, "enqueuing a null message!");
+
+  if (!_buffer->push_back(output, decorations, msg)) {
     bool p_created;
-    uint32_t* counter = _stats.add_if_absent(msg.output(), 0, &p_created);
+    uint32_t* counter = _stats.put_if_absent(output, 0, &p_created);
```
shipilev, line 77: Just checking: the rewrite `add_if_absent` -> `put_if_absent` is needed, because the underlying type changed, right?

shipilev (commented):
OK, this looks good, but I think we should fix/backport these two before backporting this one to 17u:
  https://bugs.openjdk.org/browse/JDK-8297302 (fixed, should be backported)
  https://bugs.openjdk.org/browse/JDK-8309067 (not fixed, should be dealt with first)

src/hotspot/share/logging/logAsyncWriter.cpp
navyxliu: Correct. I change from hashtable.hpp to resourceHash.hpp because the later supports arena allocator.

navyxliu:
> https://bugs.openjdk.org/browse/JDK-8309067 (not fixed, should be dealt with first)

okay.  I will work on [JDK-8309067](https://bugs.openjdk.org/browse/JDK-8309067)

PR #1639, opened 2023-08-04
8312065: Socket.connect does not timeout as expected when profiling

yanglong1010:
Hi all,

This pull request contains a fix for [JDK-8312065](https://bugs.openjdk.org/browse/JDK-8312065).

The old SocketImpl is still present in JDK 17, can be enabled with -Djdk.net.usePlainSocketImpl.

I have verified that this problem exists on Linux and macOS, and I feel that it also exists on AIX, 
so I repaired these 3 platforms. Windows implementations are different, so don't have this problem.

I added a platform dependent test.

The test needs to get the OS thread ID of the Java thread. It seems that there is no good Java API to get it directly. The minimum requirement of Linux [/proc/thread-self](https://man7.org/linux/man-pages/man5/proc.5.html#:~:text=/proc/thread%2Dself%20(since%20Linux%203.17)) is Linux 3.17.
The JFR event records the OS thread ID, although this method is cumbersome, it is reliable.

The test also needs to use the kill command to send a signal to the thread. macOS's kill seems to only support process, and I don't have an AIX environment, so this test only supports Linux.

Thanks!

### DIFF
--- a/src/java.base/aix/native/libnet/aix_close.c
+++ b/src/java.base/aix/native/libnet/aix_close.c
@@ -489,7 +489,17 @@ int NET_Connect(int s, struct sockaddr *addr, int addrlen) {
 }
 
 int NET_Poll(struct pollfd *ufds, unsigned int nfds, int timeout) {
-    BLOCKING_IO_RETURN_INT( ufds[0].fd, poll(ufds, nfds, timeout) );
+    int ret;
+    threadEntry_t self;
+    fdEntry_t *fdEntry = getFdEntry(ufds[0].fd);
+    if (fdEntry == NULL) {
+        errno = EBADF;
+        return -1;
+    }
+    startOp(fdEntry, &self);
+    ret = poll(ufds, nfds, timeout);
+    endOp(fdEntry, &self);
+    return ret;
 }
 
 /*
--- a/src/java.base/linux/native/libnet/linux_close.c
+++ b/src/java.base/linux/native/libnet/linux_close.c
@@ -396,7 +396,17 @@ int NET_Connect(int s, struct sockaddr *addr, int addrlen) {
 }
 
 int NET_Poll(struct pollfd *ufds, unsigned int nfds, int timeout) {
-    BLOCKING_IO_RETURN_INT( ufds[0].fd, poll(ufds, nfds, timeout) );
+    int ret;
+    threadEntry_t self;
+    fdEntry_t *fdEntry = getFdEntry(ufds[0].fd);
+    if (fdEntry == NULL) {
+        errno = EBADF;
+        return -1;
+    }
+    startOp(fdEntry, &self);
+    ret = poll(ufds, nfds, timeout);
+    endOp(fdEntry, &self);
+    return ret;
 }
 
 /*
--- a/test/jdk/java/net/Socket/B8312065.java
+++ b/test/jdk/java/net/Socket/B8312065.java
@@ -1,0 +1,249 @@
+/*
+ * Copyright (c) 2023, Alibaba Group Holding Limited. All Rights Reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8312065
+ * @summary Socket.connect does not timeout as expected when profiling (i.e. keep receiving signal)
+ * @library /test/lib
+ * @requires os.family == "linux"
+ * @run main/othervm/timeout=120 -Djdk.net.usePlainSocketImpl B8312065
+ */
+
+import jdk.jfr.Configuration;
+import jdk.jfr.Recording;
+import jdk.jfr.consumer.RecordedEvent;
+import jdk.jfr.consumer.RecordedThread;
+import jdk.jfr.consumer.RecordingFile;
+import sun.misc.Signal;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.Socket;
+import java.net.SocketTimeoutException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.text.ParseException;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+public class B8312065 {
+    private static volatile boolean success = false;
+    private static Recording recording = null;
+    private static volatile boolean isSendSignal = true;
+    private static volatile boolean signalReceived = false;
+    private static final boolean debug = true;
+
+    public static void main(String[] args) throws Exception {
+        // If the thread executing java.net.PlainSocketImpl.socketConnect receives continuous signals,
+        // the thread will be blocked for a long time (about 2 minutes).
+        // To test in the sub-thread, the main thread can detect the execution time of the sub-thread to avoid the
+        // test duration being too long.
+        Thread t = new Thread(() -> {
+            try {
+                task();
+            } catch (Exception e) {
+                error(e.getMessage(), e);
+            }
+        });
+        t.setDaemon(true);
+        t.start();
+
+        int timeoutSeconds = 10;
+        t.join(timeoutSeconds * 1000);
+        if (t.isAlive()) {
+            throw new RuntimeException("Test Failed: " + timeoutSeconds +
+                    " seconds have passed and it has not timed out");
+        }
+
+        if (!success) {
+            throw new RuntimeException("Test Failed");
+        }
+    }
+
+    private static void task() throws IOException, ParseException, InterruptedException {
+        Thread.currentThread().setName("B8312065");
+
+        // Find OS thread ID of the current thread
+        long osThreadId = getOSThreadId();
+        if (osThreadId == 0) {
+            throw new RuntimeException("Failed to get operating system thread id");
+        }
+
+        Thread t = startSendingSignalToThread(osThreadId);
+
+        test();
+
+        isSendSignal = false;
+        t.join();
+    }
+
+    private static Thread startSendingSignalToThread(long osThreadId) throws InterruptedException {
+        // Setup SIGPROF handler
+        Signal.handle(new Signal("PROF"), (signal) -> {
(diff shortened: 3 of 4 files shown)


test/jdk/java/net/Socket/B8312065.java
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * The JFR event records the OS thread ID, which is a reliable way.
+     * Another way is use /proc/thread-self, but unfortunately, it's not supported until linux 3.17
+     * @return Operating System Thread id
+     * @throws IOException
+     * @throws ParseException
+     */
+    private static long getOSThreadId() throws IOException, ParseException {
+        startJFR();
```
apangin, line 181: You may reuse `libNativeThread` for finding threadId and sending signals. This should make the test a lot simpler.
See `SocketAcceptInterruptTest` and `SocketReadInterruptTest` for example.
JFR is an optional feature, it is not available in Minimal VM.

test/jdk/java/net/Socket/B8312065.java
```diff
@@ -0,0 +1,249 @@
...
+
+    static void error(String message, Exception e) {
+        System.out.println(message);
+        e.printStackTrace();
+    }
+
+    static void passed() {
+        success = true;
+    }
+}
```
apangin, line 249: EOL is missing on the last line

test/jdk/java/net/Socket/B8312065.java
yanglong1010: Thanks for your advice. I have updated the test code.

yanglong1010:
Gentle ping @apangin

apangin:
@yanglong1010 I've added a minor comment, otherwise the fix looks good to me. But you'll need one more review anyway, as I don't have an official reviewer status.
Please also configure your repository to [enable pre-submit tests](https://github.com/openjdk/jdk17u-dev/pull/1639/checks?check_run_id=16167188975).

test/jdk/java/net/Socket/B8312065.java
```diff
@@ -0,0 +1,82 @@
...
+                e.printStackTrace();
+            } finally {
+                System.out.println("Test FAILED");
+                System.exit(1);
+            }
+        });
+        t.setDaemon(true);
+        t.start();
+
+        latch.await();
```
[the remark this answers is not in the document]
yanglong1010: It's really useless and has been deleted.

yanglong1010:
> @yanglong1010 I've added a minor comment, otherwise the fix looks good to me. But you'll need one more review anyway, as I don't have an official reviewer status. Please also configure your repository to [enable pre-submit tests](https://github.com/openjdk/jdk17u-dev/pull/1639/checks?check_run_id=16167188975).

Thanks a lot. I have done all your suggestions.

test/jdk/java/net/Socket/B8312065.java
```diff
@@ -0,0 +1,76 @@
...
+        });
+        t.setDaemon(true);
+        t.start();
+
+        try {
+            Socket socket = new Socket();
+            // There is no good way to mock SocketTimeoutException, just assume 192.168.255.255 is not in use
+            socket.connect(new InetSocketAddress("192.168.255.255", 8080), 2000);
+        } catch (SocketTimeoutException e) {
+            System.out.println("Test passed");
```
vyommani, line 73: i will suggest you to provide a check that test 'socket.connect' throws 'SocketTimeoutException'  only after the time expires.

store time just before socket connect and in catch block check if time spent is >= timeout specified.

test/jdk/java/net/Socket/B8312065.java
```diff
@@ -0,0 +1,76 @@
...
+                    if (NativeThread.signal(osThreadId, NativeThread.SIGPIPE) != 0) {
+                        System.out.println("Failed to send signal");
+                        System.exit(1);
+                    }
+                    Thread.sleep(1000);
+                }
+            } catch (Exception e) {
+                e.printStackTrace();
+            } finally {
+                System.out.println("Test FAILED");
```
vyommani, line 61: don't you think "Test FAILED" will print always ?

test/jdk/java/net/Socket/B8312065.java
yanglong1010: Thanks, updated.

test/jdk/java/net/Socket/B8312065.java
yanglong1010: Hi vyommani. There are two situations in which this sentence is executed, one is that it loops for 10 seconds (this will not happen on the JDK I repaired), and the other is that the sleep is interrupted (it will not happen).
I ran the test on the JDK containing my fix and watched stdout, the message was indeed not printed and the test passed. On the JDK before the repair, it will print, and the test will fail.
But in order to make this code look clearer, I removed 'finnaly'.

phohensee (changes requested):
Code duplication is the source of many bugs, so nstead of duplicating the macro code inline, I suggest refactoring a bit and add a RETRY macro argument, vis,
```
#define BLOCKING_IO_RETURN_INT(FD, FUNC, RETRY) {     \
    int ret;                                          \
    threadEntry_t self;                               \
    fdEntry_t *fdEntry = getFdEntry(FD);              \
    if (fdEntry == NULL) {                            \
        errno = EBADF;                                \
        return -1;                                    \
    }                                                 \
    do {                                              \
        startOp(fdEntry, &self);                      \
        ret = FUNC;                                   \
        endOp(fdEntry, &self);                        \
    } while ((RETRY) && ret == -1 && errno == EINTR); \
    return ret;                                       \
}
```
Then, add "true" as the last argument to all invocations of the macro, except the one in NET_Poll, which becomes
```
int NET_Poll(struct pollfd *ufds, unsigned int nfds, int timeout) {
    BLOCKING_IO_RETURN_INT( ufds[0].fd, poll(ufds, nfds, timeout), false );
}
```

phohensee:
The GHA failures seem to be infrastructure issues. Your next merge will rerun them, hopefully successfully.

src/java.base/aix/native/libnet/aix_close.c
```diff
@@ -489,7 +489,17 @@ int NET_Connect(int s, struct sockaddr *addr, int addrlen) {
...
+    threadEntry_t self;
+    fdEntry_t *fdEntry = getFdEntry(ufds[0].fd);
+    if (fdEntry == NULL) {
+        errno = EBADF;
+        return -1;
+    }
+    startOp(fdEntry, &self);
+    ret = poll(ufds, nfds, timeout);
+    endOp(fdEntry, &self);
+    return ret;
```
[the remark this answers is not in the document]
yanglong1010: hi, vyommani, thanks for the suggestion. Do you mind if I modify it like Paul did, I feel like this would make the macro cleaner.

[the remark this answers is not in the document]
vyommani: yes, that's more cleaner approach.

src/java.base/linux/native/libnet/linux_close.c
```diff
@@ -40,6 +40,13 @@
 #include "jvm.h"
 #include "net_util.h"
 
+#if (__STDC_VERSION__ >= 199901L)
+#include <stdbool.h>
+#else
+#define true 1
+#define false 0
+#endif
+
```
apangin, line 49: Since the file already includes `jvm.h`, you can use `JNI_TRUE`/`JNI_FALSE` to avoid introducing more ifdefs

yanglong1010: Andrei, thanks. Pushed.

yanglong1010:
> Code duplication is the source of many bugs, so instead of duplicating the macro code inline, add a RETRY macro argument, vis,
> 
> ```
> #define false 0
> #define true 1
> 
> #define BLOCKING_IO_RETURN_INT(FD, FUNC, RETRY) {     \
>     int ret;                                          \
>     threadEntry_t self;                               \
>     fdEntry_t *fdEntry = getFdEntry(FD);              \
>     if (fdEntry == NULL) {                            \
>         errno = EBADF;                                \
>         return -1;                                    \
>     }                                                 \
>     do {                                              \
>         startOp(fdEntry, &self);                      \
>         ret = FUNC;                                   \
>         endOp(fdEntry, &self);                        \
>     } while ((RETRY) && ret == -1 && errno == EINTR); \
>     return ret;                                       \
> }
> ```
> 
> Then, add "true" as the last argument to all invocations of the macro, except the one in NET_Poll, which becomes
> 
> ```
> int NET_Poll(struct pollfd *ufds, unsigned int nfds, int timeout) {
>     BLOCKING_IO_RETURN_INT( ufds[0].fd, poll(ufds, nfds, timeout), false );
> }
> ```

Paul, Thank you very much for your detailed advice.

test/jdk/java/net/Socket/B8312065.java
```diff
@@ -0,0 +1,87 @@
...
+                }
+            }
+            System.out.println("Test FAILED: Socket.connect blocked " + n + " seconds, " +
+                    "expected around " + timeout / 1000 + " seconds");
+            System.exit(1);
+        });
+        t.setDaemon(true);
+        t.start();
+
+        long startMillis = System.currentTimeMillis();
```
vyommani, line 71: please use monotonically increasing time(System.nanoTime()) in place of 'currentTimeMillis'

vyommani (commented):
There are couples of formatting error(green bar), i think it is because of tab. Please use space in place of tab.

test/jdk/java/net/Socket/B8312065.java
yanglong1010: Thanks, vyommani.  I have pushed. System wall-clock may be changed at any time, It is indeed more reasonable to use `nanoTime`.

phohensee (approved):

yanglong1010:
> There are couples of formatting error(green bar), i think it is because of tab. Please use space in place of tab.

vyommani, I carefully checked the files I have modified, and I did not find `tab`s, all the blanks are `space`s.
I think it may be because the line where the `while` keyword is located has become longer. For alignment, I added spaces in adjacent lines.

vyommani (approved):
Looks OK to me.

apangin (approved):

GoeLin:
Could you please trigger the GHA tests again and the supply a Fix request comment.
I remove the tag in the meantime.

yanglong1010:
> Could you please trigger the GHA tests again and the supply a Fix request comment. I remove the tag in the meantime.

Of course, my pleasure.

apangin:
@yanglong1010 I guess you need to merge the latest changes from master to fix GitHub Actions.

GoeLin:
You are still missing the fix request comment that states reason, risk and testing in the JBS bug.
You can issue another approval request to do that, or edit JBS.  
For example see https://bugs.openjdk.org/browse/JDK-8232933

yanglong1010:
/approval 8312065 request I would like to fix this issue. 
This problem can be reproduced using pure Java code, or it can be reproduced by sending signal from an external profiling tool. If triggered, it may cause the thread to block abnormally. 
Low risk: The repair is clear，do not perform redundant loop retries for NET_Poll. Existing tests have proven that it does not affect the original functions, and a dedicated test for this issue has been added.
Testing: All GHA tests are passed. I ran all the jtreg tests (test/jdk/java/net/ServerSocket,test/jdk/java/net/Socket,test/jdk/java/net/SocketOption,test/jdk/java/net/ipv6) related to jdk.net.usePlainSocketImpl on my host, all passed. I've also added a test for this issue, the unfixed JDK fails, the fixed version passes.

yanglong1010:
> You are still missing the fix request comment that states reason, risk and testing in the JBS bug. You can issue another approval request (Use: approval request " to do that, or edit JBS. For example see https://bugs.openjdk.org/browse/JDK-8232933

GoeLin, Thanks. Updated.

D-D-H:
@yanglong1010 I'm not a committer of jdk-updates project :-(

yanglong1010:
Thanks a lot, @apangin @vyommani @phohensee @GoeLin.

PR #1672, opened 2023-08-17
8314552: Fix javadoc tests to work with jtreg 7

lutkerd:
Updating the tests to specify a UTF-8 encoding so that they work with jtreg-7 and the embedded testng jar.

```
bash ./configure --with-jtreg=jtreg-7.3+1
make images
make run-test TEST="jtreg:test/langtools/jdk/javadoc/doclet"
```

All tests pass.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8314552](https://bugs.openjdk.org/browse/JDK-8314552): Fix javadoc tests to work with jtreg 7 (**Bug** - P4)

### Reviewers
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**)
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - **Reviewer**)
 * [Andrew John Hughes](https://openjdk.org/census#andrew) (@gnu-andrew - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/1672/head:pull/1672` \
`$ git checkout pull/1672`

Update a local copy of the PR: \
`$ git checkout pull/1672` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/1672/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1672`

View PR using the GUI difftool: \
`$ git pr show -t 1672`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk17u-dev/pull/1672.diff">https://git.openjdk.org/jdk17u-dev/pull/1672.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/1672#issuecomment-1683122485)

PR #1595, opened 2023-07-18
8289798: Update to use jtreg 7

GoeLin:
I want to backport this on the path to jtreg 7.2.

I had to move the edit of lib-tests.m4 to toolchain.m4

All the others I resolved because we skipped the step to 6.1.

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8289798](https://bugs.openjdk.org/browse/JDK-8289798): Update to use jtreg 7 (**Enhancement** - P3)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/1595/head:pull/1595` \
`$ git checkout pull/1595`

Update a local copy of the PR: \
`$ git checkout pull/1595` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/1595/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1595`

View PR using the GUI difftool: \
`$ git pr show -t 1595`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk17u-dev/pull/1595.diff">https://git.openjdk.org/jdk17u-dev/pull/1595.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/1595#issuecomment-1640207753)

PR #1673, opened 2023-08-18
8284772: GHA: Use GCC Major Version Dependencies Only

gnu-andrew:
GHA regularly breaks because we specify a very explicit GCC version, even down to the release versioning of the Ubuntu package.

Rather than bumping this yet again like https://github.com/openjdk/jdk17u/commit/260f28761a87553fd634615df5ba3db5b76ae692, this PR suggests dropping the specific version as we did some time ago in 8u and I've also proposed to fix the same breakage in 11u (https://github.com/openjdk/jdk11u-dev/pull/2087). The requirement still specifies a specific major version of GCC. It just means the dependency isn't broken every time Ubuntu bumps to a new minor release or even just makes a minor change to the package alone.

Note that the current setup does not guarantee sticking with an exact version of GCC anyway, because - as seen by recent GHA breakage - older versions get removed from the package repository. All we get from this exact version requirement is sporadic breakage. If we truly want a static version of GCC, we need to provide our own as we do with the JDK.

~~This is based on https://github.com/openjdk/jdk17u-dev/pull/1595 to allow https://github.com/openjdk/jdk17u-dev/pull/1672 to proceed.~~ This is now based on master and the change is in trunk.

---------
### Progress

### Issue
 * [JDK-8284772](https://bugs.openjdk.org/browse/JDK-8284772): GHA: Use GCC Major Version Dependencies Only (**Bug** - P4)

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/1673/head:pull/1673` \
`$ git checkout pull/1673`

`$ git checkout pull/1673` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/1673/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1673`

`$ git pr show -t 1673`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/1673.diff">https://git.openjdk.org/jdk17u-dev/pull/1673.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/1673#issuecomment-1684151248)

### DIFF
--- a/.github/workflows/build-cross-compile.yml
+++ b/.github/workflows/build-cross-compile.yml
@@ -31,12 +31,6 @@ on:
       gcc-major-version:
         required: true
         type: string
-      apt-gcc-version:
-        required: true
-        type: string
-      apt-gcc-cross-version:
-        required: true
-        type: string
       extra-conf-options:
         required: false
         type: string
@@ -113,10 +107,10 @@ jobs:
           sudo apt-get update
           sudo apt-get install --only-upgrade apt
           sudo apt-get install \
-              gcc-${{ inputs.gcc-major-version }}=${{ inputs.apt-gcc-version }} \
-              g++-${{ inputs.gcc-major-version }}=${{ inputs.apt-gcc-version }} \
-              gcc-${{ inputs.gcc-major-version }}-${{ matrix.gnu-arch }}-linux-gnu${{ matrix.gnu-abi}}=${{ inputs.apt-gcc-cross-version }} \
-              g++-${{ inputs.gcc-major-version }}-${{ matrix.gnu-arch }}-linux-gnu${{ matrix.gnu-abi}}=${{ inputs.apt-gcc-cross-version }} \
+              gcc-${{ inputs.gcc-major-version }} \
+              g++-${{ inputs.gcc-major-version }} \
+              gcc-${{ inputs.gcc-major-version }}-${{ matrix.gnu-arch }}-linux-gnu${{ matrix.gnu-abi}} \
+              g++-${{ inputs.gcc-major-version }}-${{ matrix.gnu-arch }}-linux-gnu${{ matrix.gnu-abi}} \
               libxrandr-dev libxtst-dev libcups2-dev libasound2-dev
           sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-${{ inputs.gcc-major-version }} 100 --slave /usr/bin/g++ g++ /usr/bin/g++-${{ inputs.gcc-major-version }}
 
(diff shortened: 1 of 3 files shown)


shipilev (changes requested):
Nope, not for 17u.

While the actual removal of minor GCC version tracking might be okay, it is wrong to keep effectively _forward-porting_ the GHA change from 8u -> 11u -> 17u. You need to do this in mainline, argue the point on minor GCC version removals (in other words, the reversal of [JDK-8256393](https://bugs.openjdk.org/browse/JDK-8256393)), and then argue for backports. Otherwise, you will keep doing this for every subsequent XXu.

tstuefe:
I'm very much in favor of dropping the GCC minor version dependency since it has been a huge pain. I don't care for patch direction, as long as we can convince mainline to do the same.

gnu-andrew:
> Nope, not for 17u.
> 
> While the actual removal of minor GCC version tracking might be okay, it is wrong to keep effectively _forward-porting_ the GHA change from 8u -> 11u -> 17u. You need to do this in mainline, argue the point on minor GCC version removals (in other words, the reversal of [JDK-8256393](https://bugs.openjdk.org/browse/JDK-8256393)), and then argue for backports. Otherwise, you will keep doing this for every subsequent XXu.

Yes, this was never intended to be merged in the current state anyway. It was provided so we could see if #1672 would actually fix the JTReg failures.

My intention was to propose it to trunk as well in due course and I've now done that (https://github.com/openjdk/jdk/pull/15374). We can still use this PR for the subsequent 17u PR. I would very much like to see someone argue in favour of keeping this versioning. It seems to have been added with little review and has baffled me from the start. I was surprised to find it was still present over this last week. I had hoped someone else would have ditched it long ago. I tend to be focused mostly on 8u, where we dropped it from the start and haven't had any problems since.

gnu-andrew:
> I'm very much in favor of dropping the GCC minor version dependency since it has been a huge pain. I don't care for patch direction, as long as we can convince mainline to do the same.

Hopefully. Having had tests fail three times in the last week due to this (once in 11u, once in 17u and once in our own fork of 21u), I'm surprised people have stuck with it this long.

shipilev (approved):
Looks okay now. I guess we should be doing these in every maintained release at this point.

shipilev:
@gnu-andrew, you need `jdk17u-fix-request` on it now, I think.

gnu-andrew:
> Looks okay now. I guess we should be doing these in every maintained release at this point.

Yes, I should open a PR for 21u. Added `jdk17u-fix-request` for this one now. Thanks for the review.

gnu-andrew:
I see `jdk17u-fix-yes`
/integrate

PR #1679, opened 2023-08-21
8312182: THPs cause huge RSS due to thread start timing issue

tstuefe:
Unclean composite backport to jdk17u. Fixes JDK-8312182 - "THPs cause huge RSS due to thread start timing issue" (https://bugs.openjdk.org/browse/JDK-8312182)

Problem:

On a machine with transparent huge pages (THP) unconditionally enabled (/sys/kernel/mm/transparent_hugepage/enabled = "always"), the JVM may show a huge memory footprint (RSS) and degraded thread start performance.

The following factors make the problem more severe and more likely:
- thread stack size of 2M (on arm64 or x64) or larger
- many threads, or high thread creation churn
- a slow or overloaded machine (since part of the problem is timing-dependent)

For a detailed discussion of the underlying problem, please see https://github.com/openjdk/jdk/pull/14919.

----------------

In jdk Head, the issue got fixed with a sequence of patches:

- [JDK-8303215](https://bugs.openjdk.org/browse/JDK-8303215) "Make thread stacks not use huge pages"
- [JDK-8312182](https://bugs.openjdk.org/browse/JDK-8312182) "THPs cause huge RSS due to thread start timing"

However, JDK-8312182 itself needed one preparatory fix:
- [JDK-8310233](https://bugs.openjdk.org/browse/JDK-8310233) "Fix THP detection on Linux"

and then we had several corner-case test problems which are fixed with:
- [JDK-8312394](https://bugs.openjdk.org/browse/JDK-8312394) "[linux] SIGSEGV if kernel was built without hugepage support"
- [JDK-8312620](https://bugs.openjdk.org/browse/JDK-8312620) "WSL Linux build crashes after JDK-8310233"
- [JDK-8314139](https://bugs.openjdk.org/browse/JDK-8314139) "TEST_BUG: runtime/os/THPsInThreadStackPreventionTest.java could fail on machine with large number of cores"

and finally, we decided to rename the switch that allows to switch off the THP mitigation with a final patch:
- [JDK-8312585](https://bugs.openjdk.org/browse/JDK-8312585) "Rename DisableTHPStackMitigation flag to THPStackMitigation"

Instead of downporting these 7 patches verbatim, I prepared a composite patch containing only the necessary mitigation and mitigation tests.

This is similar to the [jdk11u downport](https://github.com/openjdk/jdk11u-dev/pull/2086), but in jdk17u, [JDK-8303215](https://bugs.openjdk.org/browse/JDK-8303215) had been already backported. Therefore there are some minor differences.

This patch does:
- make sure that all thread stacks have at least one glibc guard page to prevent clustering of adjacent thread stacks into one VMA
- change the default size of stacks to be not aligned to 2MB to prevent intra-stack THPs from forming

The patch needs some infrastructure, but I downported only the necessary parts: the helper class "HugePages", which is used in head to scan the operating system for information about THP settings. I only included the parts to do with THPs and left the rest out.

The patch also includes a regression test.

------

Testing:

I manually tested the JVM on Linux x64 with THP=always:

Without the patch (-Xmx1g -Xms1g -XX:+AlwaysPreTouch -Xss2m, 10000 threads started), I see slow thread startup and *11 GB - 14 GB* of RSS.

The patched version comes up a lot faster and only shows *1.3* GB of RSS.

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))

### Issues
 * [JDK-8312182](https://bugs.openjdk.org/browse/JDK-8312182): THPs cause huge RSS due to thread start timing issue (**Bug** - P3)
 * [JDK-8310687](https://bugs.openjdk.org/browse/JDK-8310687): JDK-8303215 is incomplete (**Bug** - P4)
 * [JDK-8310233](https://bugs.openjdk.org/browse/JDK-8310233): Fix THP detection on Linux (**Bug** - P4)
 * [JDK-8312394](https://bugs.openjdk.org/browse/JDK-8312394): [linux] SIGSEGV if kernel was built without hugepage support (**Bug** - P3)
 * [JDK-8312620](https://bugs.openjdk.org/browse/JDK-8312620): WSL Linux build crashes after JDK-8310233 (**Bug** - P3)
 * [JDK-8314139](https://bugs.openjdk.org/browse/JDK-8314139): TEST_BUG: runtime/os/THPsInThreadStackPreventionTest.java could fail on machine with large number of cores (**Bug** - P4)
 * [JDK-8312585](https://bugs.openjdk.org/browse/JDK-8312585): Rename DisableTHPStackMitigation flag to THPStackMitigation (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/1679/head:pull/1679` \
`$ git checkout pull/1679`

`$ git checkout pull/1679` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/1679/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1679`

`$ git pr show -t 1679`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/1679.diff">https://git.openjdk.org/jdk17u-dev/pull/1679.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/1679#issuecomment-1686620890)

### DIFF
--- a/src/hotspot/os/linux/globals_linux.hpp
+++ b/src/hotspot/os/linux/globals_linux.hpp
@@ -85,6 +85,15 @@
   product(bool, UseCpuAllocPath, false, DIAGNOSTIC,                     \
           "Use CPU_ALLOC code path in os::active_processor_count ")     \
                                                                         \
+  product(bool, THPStackMitigation, true, DIAGNOSTIC,                   \
+          "If THPs are unconditionally enabled on the system (mode "    \
+          "\"always\"), the JVM will prevent THP from forming in "      \
+          "thread stacks. When disabled, the absence of this mitigation"\
+          "allows THPs to form in thread stacks.")                      \
+                                                                        \
+  develop(bool, DelayThreadStartALot, false,                            \
+          "Artificially delay thread starts randomly for testing.")     \
+                                                                        \
   product(bool, DumpPerfMapAtExit, false, DIAGNOSTIC,                   \
           "Write map file for Linux perf tool at exit")                 \
                                                                         \
--- a/src/hotspot/os/linux/hugepages.cpp
+++ b/src/hotspot/os/linux/hugepages.cpp
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) 2005, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2023, Red Hat Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+#include "precompiled.hpp"
+#include "hugepages.hpp"
+
+#include "logging/log.hpp"
+#include "logging/logStream.hpp"
+#include "runtime/os.hpp"
+#include "utilities/debug.hpp"
+#include "utilities/globalDefinitions.hpp"
+#include "utilities/ostream.hpp"
+
+// Given a file that contains a single (integral) number, return that number in (*out) and true;
+// in case of an error, return false.
+static bool read_number_file(const char* file, size_t* out) {
+  FILE* f = ::fopen(file, "r");
+  bool rc = false;
+  if (f != NULL) {
+    uint64_t i = 0;
+    if (::fscanf(f, SIZE_FORMAT, out) == 1) {
+      rc = true;
+    }
+    ::fclose(f);
+  }
+  return rc;
+}
+
+THPSupport::THPSupport() :
+    _initialized(false), _mode(thp_never), _pagesize(SIZE_MAX) {}
+
+
+THPMode THPSupport::mode() const {
+  assert(_initialized, "Not initialized");
+  return _mode;
+}
+
+size_t THPSupport::pagesize() const {
+  assert(_initialized, "Not initialized");
+  return _pagesize;
+}
+
+void THPSupport::scan_os() {
+  // Scan /sys/kernel/mm/transparent_hugepage/enabled
+  // see mm/huge_memory.c
+  _mode = thp_never;
+  const char* filename = "/sys/kernel/mm/transparent_hugepage/enabled";
+  FILE* f = ::fopen(filename, "r");
+  if (f != NULL) {
+    char buf[64];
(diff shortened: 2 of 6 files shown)


shipilev:
Process-wise and for historical record, if you do a composite patch out of several issues, you need at least mark those issues as fixed by this PR. Tell bot like this: `/issue add 8303215,8312182,...`

tstuefe:
/issue add  8310233 8312394 8312620 8314139 8312585

tstuefe:
Friendly ping. It would be good to get this fixed in time for the next CPU.

shipilev:
> Friendly ping. It would be good to get this fixed in time for the next CPU.

I have concerns about this, for several reasons:
 a) The fix has a considerable bugtail _in the product code_, which does not inspire confidence there would not be more soon;
 b) The fix is enabled by default, which means any bugs in it would be exposed to users immediately;
 c) The review is hard, because it cobbles 7 commits together, so it would take a while to disentangle if any mistakes crept in (note: the usual way to deal with this problem is to propose the PR with a commit per issue, so we can review individual commits);

This also does not look like a recent regression, but rather a long-standing bug, right?
So, is there a reason to rush it for 17.0.9 in October?

tstuefe:
@shipilev Thanks for looking at this. I debated with myself for a long time whether this was the right approach. I did not choose to build a composite patch out of laziness (if anything, downporting the issues separately and verbatim would have been much simpler, albeit slower).

By providing a minimal (not cobbled together but carefully selected) patch I minimize the problem surface because I leave out code that have nothing to do with the goal of this patch: the static hugepage detection of [JDK-8310233](https://bugs.openjdk.org/browse/JDK-8310233) "Fix THP detection on Linux").

> * The fix has a considerable bugtail _in the product code_, which does not inspire confidence there will not be more soon;

we have two trailing bugs in product code:
a) [JDK-8312394](https://bugs.openjdk.org/browse/JDK-8312394) "[linux] SIGSEGV if kernel was built without hugepage support"
b) [JDK-8312620](https://bugs.openjdk.org/browse/JDK-8312620) "WSL Linux build crashes after JDK-8310233"

(a) JDK-8312394 is of no concern since it only affects code I explicitly left out from the patch (it does not affect THPs). It would be a concern were I to downport patches individually and verbatim.

That leaves (b), which is a bug in a super obscure context (building on Windows with WSL that carries an arguably broken Linux kernel). One bug is not a very long tail.

> 
> * The fix is enabled by default, which means any bugs in it would be exposed to users immediately;
> 

I think this patch is just not that risky.

> * The review is hard, because it cobbles 7 commits together, so it would take a while to disentangle if any mistakes crept in (note: the usual way to deal with this problem is to propose the PR with a commit per issue, so we can review individual commits);

But that process would carry more risk since I would have to downport unnecessary parts and have time windows with unfixed bugs.

This also shows a shortcoming of the review process. If I downport stuff verbatim, reviews are simple since they are mostly mechanical mental diffing; but that is not the most ideal patch, which would be one that is small and confined.

> 
> This also does not look like a recent regression, but rather a long-standing bug, right? So, is there a reason to rush it for 17.0.9 in October?

We have customers running with THP enabled always and unwilling or unable to change that; they would be happy about a fix. 

I'm fine with postponing this patch (not that I have any choice since I lack reviews and the window is almost closed). But the whole discussion leaves me dissatisfied with the practice of downporting whole patch trees to get a single issue fixed. We recently had similar discussions when downporting https://github.com/openjdk/jdk11u-dev/pull/2035 which ended up a far bigger change than necessary, carrying a lot of code for the sole purpose of keeping a low delta to upstream.

tstuefe:
okay, I'll withdraw. Let's do it piece by piece as usual.

shipilev:
> okay, I'll withdraw. Let's do it piece by piece as usual.

Thanks!

Yes, piece by piece would be the right approach here. I don't actually mind clustering several changes into one PR, as long as PR commits tell the story well: what was picked, in what order, and what changes were done along the way.

I also don't mind bringing in safe-ish improvements to the common code if it resolves significant deviation from mainline. A palpable number of deviations we did over the years bit us in the back at unfortunate times...

PR #1857, opened 2023-10-08
Backport 6b942893868fa1a64977288bdbdb1bbff8bd9d9c

omikhaltsova:
I'd like to backport JDK-8309502 to jdk17u. This fix is related only to RISC-V.
The patch applies almost cleanly except `src/hotspot/cpu/riscv/c2_MacroAssembler_riscv.cpp` because of deletion an empty line at the end in the original patch.


VladimirKempik (approved):
lgtm

PR #2039, opened 2023-12-11
8317507: C2 compilation fails with "Exceeded _node_regs array"

shipilev:
Clean backport to fix the C2 corner case.

Additional testing:
 - [x] New regression test fails without the fix, passes with it
 - [x] Linux x86_64 server fastdebug,`tier{1,2,3,4}`
 - [x] Linux AArch64 server fastdebug,`tier{1,2,3}`

---------
### Progress
- [x] [JDK-8317507](https://bugs.openjdk.org/browse/JDK-8317507) needs maintainer approval

### Issue
 * [JDK-8317507](https://bugs.openjdk.org/browse/JDK-8317507): C2 compilation fails with "Exceeded _node_regs array" (**Bug** - P2 - Approved)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/2039/head:pull/2039` \
`$ git checkout pull/2039`

`$ git checkout pull/2039` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/2039/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2039`

`$ git pr show -t 2039`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/2039.diff">https://git.openjdk.org/jdk17u-dev/pull/2039.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/2039#issuecomment-1851753858)

### DIFF
--- a/src/hotspot/share/adlc/output_c.cpp
+++ b/src/hotspot/share/adlc/output_c.cpp
@@ -3034,6 +3034,9 @@ static void define_fill_new_machnode(bool used, FILE *fp_cpp) {
     fprintf(fp_cpp, "    if( i != cisc_operand() ) \n");
     fprintf(fp_cpp, "      to[i] = _opnds[i]->clone();\n");
     fprintf(fp_cpp, "  }\n");
+    fprintf(fp_cpp, "  // Do not increment node index counter, since node reuses my index\n");
+    fprintf(fp_cpp, "  Compile* C = Compile::current();\n");
+    fprintf(fp_cpp, "  C->set_unique(C->unique() - 1);\n");
     fprintf(fp_cpp, "}\n");
   }
   fprintf(fp_cpp, "\n");
--- a/test/hotspot/jtreg/compiler/regalloc/TestNodeRegArrayOverflow.java
+++ b/test/hotspot/jtreg/compiler/regalloc/TestNodeRegArrayOverflow.java
@@ -1,0 +1,599 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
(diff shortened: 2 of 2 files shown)


shipilev:
/approval request Clean backport to fix the C2 corner case. Applies cleanly. Tests pass.

GoeLin:
Won't this increase the memory consumption of the compiler significantly? As I understand it can push more nodes on the arenas now.

shipilev:
> Won't this increase the memory consumption of the compiler significantly? As I understand it can push more nodes on the arenas now.

The footprint increase effect might be in that we would be able to carry more nodes in these degenerate situations before bailing. Currently it is capped by `PhaseRegAlloc::alloc_node_regs`, which is a small number: original report mentions 1118 nodes for the test case in question. But AFAIU, it is basically a choice what do you do if compiler/user over-unrolled: before this patch, C2 would eventually just bail; after this patch, C2 would still compile the thing.

GoeLin:
Yes, it will still compile the method. But it might need twice as many memory as before. This might kill existing installations. Compiler memory is often a bottle neck.

shipilev:
Yeah, accepting more nodes is a disadvantage of the fix. But if we look closely at the way it fails in fastdebug for the given test cases, I think assert catches the OOB access in `PhaseCFG::insert_goto_at` calling to `PhaseRegAlloc::set_bad`, which suggests that we are actually dealing with memory stomp, not a clean bail. This explains why bug is P2. So, on balance, accepting more compiler footprint to avoid compiler memory stomp is good?

shipilev:
@GoeLin, I see `jdk17u-fix-yes` now. I take it as the signal that you are okay with the explanation above, and think this makes sense for jdk17u-dev? Just checking.

GoeLin:
yes :)

shipilev:
All right then!

/integrate

PR #2205, opened 2024-02-12
Backport 26755a968665545a151adce79a5227c79724bb6b

zzambers:
Backport fixing several tables in `doc/building.md`, so they would display correctly in GitHub. Change set applied almost cleanly to `doc/building.md` (only last table has one more target in original changeset). File `doc/building.html` was regenerated.

### DIFF
--- a/doc/building.html
+++ b/doc/building.html
@@ -11,14 +11,14 @@
       span.underline{text-decoration: underline;}
       div.column{display: inline-block; vertical-align: top; width: 50%;}
   </style>
-  <link rel="stylesheet" href="../make/data/docs-resources/resources/jdk-default.css" />
+  <link rel="stylesheet" href="../make/data/docs-resources/resources/jdk-default.css">
   <!--[if lt IE 9]>
     <script src="//cdnjs.cloudflare.com/ajax/libs/html5shiv/3.7.3/html5shiv-printshiv.min.js"></script>
   <![endif]-->
   <style type="text/css">pre, code, tt { color: #1d6ae5; }</style>
 </head>
 <body>
-<header id="title-block-header">
+<header>
 <h1 class="title">Building the JDK</h1>
 </header>
 <nav id="TOC">
@@ -174,22 +174,22 @@
 <table>
 <thead>
 <tr class="header">
-<th style="text-align: left;">Operating system</th>
-<th style="text-align: left;">Vendor/version used</th>
+<th>Operating system</th>
+<th>Vendor/version used</th>
 </tr>
 </thead>
 <tbody>
 <tr class="odd">
-<td style="text-align: left;">Linux</td>
-<td style="text-align: left;">Oracle Enterprise Linux 6.4 / 7.6</td>
+<td>Linux</td>
+<td>Oracle Enterprise Linux 6.4 / 7.6</td>
 </tr>
 <tr class="even">
-<td style="text-align: left;">macOS</td>
-<td style="text-align: left;">Mac OS X 10.13 (High Sierra)</td>
+<td>macOS</td>
+<td>Mac OS X 10.13 (High Sierra)</td>
 </tr>
 <tr class="odd">
-<td style="text-align: left;">Windows</td>
-<td style="text-align: left;">Windows Server 2012 R2</td>
+<td>Windows</td>
+<td>Windows Server 2012 R2</td>
 </tr>
 </tbody>
 </table>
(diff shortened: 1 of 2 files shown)


gnu-andrew (commented):
`building.md` changes look fine.
How was the HTML re-generated? I'm seeing some unrelated changes which I guess may be down to the tool version used?
~~~~
+@@ -11,14 +11,14 @@
+       span.underline{text-decoration: underline;}
+       div.column{display: inline-block; vertical-align: top; width: 50%;}
+   </style>
+-  <link rel="stylesheet" href="../make/data/docs-resources/resources/jdk-default.css" />
++  <link rel="stylesheet" href="../make/data/docs-resources/resources/jdk-default.css">
+   <!--[if lt IE 9]>
+     <script src="//cdnjs.cloudflare.com/ajax/libs/html5shiv/3.7.3/html5shiv-printshiv.min.js"></script>
+   <![endif]-->
+   <style type="text/css">pre, code, tt { color: #1d6ae5; }</style>
+ </head>
+ <body>
+-<header id="title-block-header">
++<header>
+ <h1 class="title">Building the JDK</h1>
+ </header>
+ <nav id="TOC">
~~~~

zzambers:
@gnu-andrew I have used appropriate makefile target on rhel-8 (pandoc 2.0.6)

zzambers:
I can try with different version of pandoc.

zzambers:
@gnu-andrew I have tried on few more systems, but pandoc from rhel-8 is closest to orignal generator I could get.
- [output of newer pandoc on f38](https://github.com/zzambers/jdk17u-dev/commit/tables-fix-f38) (2.19.2-21.fc38) or even ubuntu-20.04 (2.5-3build2) were much more different
- output of older pandoc on rhel-7 (1.12.3.1-2.el7) was also much more different

Unfortunately generated html does not include information about pandoc used. Rhel-8's pandoc (2.0.6-6.el8_6) seems pretty close.

gnu-andrew:
Thanks for looking into this, @zzambers. Yes, this is the nightmare any time we have checked-in generated files. I'm mulling over backporting the change to remove the generated configure in the next cycle, as that keeps being altered with unrelated changes when people using a different autoconf.

I'm happy for this to go in as is. It would be nice to have known what was used to generate the original, but I imagine it was close to what you have on RHEL 8.

gnu-andrew (approved):
Good to go.

zzambers:
/approval request Fixes format of tables in doc/building.md, so they display correctly on GitHub. (docs only change)

zzambers:
> Thanks for looking into this, @zzambers. Yes, this is the nightmare any time we have checked-in generated files. I'm mulling over backporting the change to remove the generated configure in the next cycle, as that keeps being altered with unrelated changes when people using a different autoconf.
> 
> I'm happy for this to go in as is. It would be nice to have known what was used to generate the original, but I imagine it was close to what you have on RHEL 8.

Based on my experiments and [release info](https://pandoc.org/releases.html), pandoc used was `>=2.2.1` and `<2.5` (unless it was patched), since it had:

Since it has:
`Make HTML template polyglot ([#4606](https://github.com/jgm/pandoc/issues/4606), OvidiusCicero), by making <link rel="stylesheet" href="$css$"> self-closing.`

but does not have:
`Text.Pandoc.Writers.HTML: Use plain " instead of &quot; outside of attributes.`

zzambers:
I thought maybe [debian buster (oldoldstable)](https://packages.debian.org/search?keywords=pandoc), but no. This one has `<!--[if lt IE 9]> ...` stuff missing (probably custom patch/backport, as even newer pandoc on f38 has this).

Output of pandoc on RHEL-8 seems closest (of all current pandoc packages on any version of RHEL, Debian, Ubuntu and supported Fedora).

gnu-andrew:
> I thought maybe [debian buster (oldoldstable)](https://packages.debian.org/search?keywords=pandoc), but no. This one has `<!--[if lt IE 9]> ...` stuff missing (probably custom patch/backport, as even newer pandoc on f38 has this).
> 
> Output of pandoc on RHEL-8 seems closest (of all current pandoc packages on any version of RHEL, Debian, Ubuntu and supported Fedora).

Yes it may not even be a packaged version but a local install. Thanks for investigating so thoroughly.

zzambers:
@gnu-andrew  thanks
/integrate

gnu-andrew:
> Oh, when moving on to jdk11u-dev, I have found there is a [script](https://github.com/openjdk/jdk17u-dev/blob/master/make/devkit/createPandocBundle.sh) for creating pandoc bundle in devkit. This seems to be the right pandoc. I'll try to update html as changeset is not yet in (as I am not commiter here). Hopefully bot will not be too angry about that.. :)

Oops, sorry I sponsored it before I saw this. I don't think it's too much of an issue. At least we now know the 'correct' way (though it's  hardly obvious!)

PR #2235, opened 2024-02-26
8280377: MethodHandleProxies does not correctly invoke default methods with varags

Delawen:
This is a backport of https://bugs.openjdk.org/browse/JDK-8280377 MethodHandleProxies does not correctly invoke default methods with varags

I applied the same fix that was applied to version 19 in https://github.com/openjdk/jdk/commit/a183bfb436a7dd998e602c2d16486e88c390fca1

---------
### Progress
- [x] [JDK-8280377](https://bugs.openjdk.org/browse/JDK-8280377) needs maintainer approval

### Issue
 * [JDK-8280377](https://bugs.openjdk.org/browse/JDK-8280377): MethodHandleProxies does not correctly invoke default methods with varags (**Bug** - P4 - Approved)

### Reviewers
 * [Andrew Haley](https://openjdk.org/census#aph) (@theRealAph - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/2235/head:pull/2235` \
`$ git checkout pull/2235`

`$ git checkout pull/2235` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/2235/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2235`

`$ git pr show -t 2235`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/2235.diff">https://git.openjdk.org/jdk17u-dev/pull/2235.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/2235#issuecomment-1964087062)


Delawen:
/approval JDK-8280377 request Fix Request
This fix is a backport. I am applying the same fix that was done for version 19
It includes tests to make sure that when we call MethodHandleProxies, it does correctly invoke default methods both with and without varags

GoeLin:
Hi @Delawen, I think this is a larger change that should get some more testing than just checking that is fixes the issue.

Delawen:
I tested with the `Main.java` example that is provided in the bug:

```
import java.lang.invoke.MethodHandleProxies;
import java.lang.invoke.MethodHandles;

import static java.lang.invoke.MethodType.methodType;

public class Main {

    public interface Interface
    {
        void method(Object... args);
        default void defaultMethod(Object... args) {
            method(args);
        }
        default void defaultMethod2(Object[] args) {
            method(args);
        }
    }

    private static void callMeViaMethodHandles(String foo, int bar) {
        System.out.println("foo=" + foo + ", bar=" + bar);
    }

    public static void main(String[] args) throws NoSuchMethodException, IllegalAccessException
    {
        var mh = MethodHandles.lookup().findStatic(
                Main.class, "callMeViaMethodHandles", methodType(void.class, String.class, int.class)
        );
        mh = mh.asSpreader(Object[].class, mh.type().parameterCount());
        var proxy = MethodHandleProxies.asInterfaceInstance(Interface.class, mh);

        // throws incorrect exception
        proxy.defaultMethod("should be foo", 3);
        
        // works fine
        proxy.defaultMethod2(new Object[] { "should be foo", 3 });
    }
} 
```
With the regular Java 17 installed on my Debian, I get the expected error pre-fix:

```
$ javac Main.java ; java --version; java Main
openjdk 17.0.10 2024-01-16
OpenJDK Runtime Environment (build 17.0.10+7-Debian-1deb12u1)
OpenJDK 64-Bit Server VM (build 17.0.10+7-Debian-1deb12u1, mixed mode, sharing)
Exception in thread "main" java.lang.IllegalArgumentException: array is not of length 2
	at java.base/java.lang.invoke.MethodHandleStatics.newIllegalArgumentException(MethodHandleStatics.java:167)
	at java.base/java.lang.invoke.MethodHandleImpl.checkSpreadArgument(MethodHandleImpl.java:584)
	at java.base/java.lang.invoke.MethodHandleProxies$1.invoke(MethodHandleProxies.java:198)
	at jdk.proxy1/com.sun.proxy.jdk.proxy1.$Proxy0.method(Unknown Source)
	at Main$Interface.defaultMethod(Main.java:12)
	at java.base/java.lang.invoke.MethodHandleProxies.callDefaultMethod(MethodHandleProxies.java:354)
	at java.base/java.lang.invoke.MethodHandleProxies$1.invoke(MethodHandleProxies.java:205)
	at jdk.proxy1/com.sun.proxy.jdk.proxy1.$Proxy0.defaultMethod(Unknown Source)
	at Main.main(Main.java:32)
```

After fixing it, I get the correct answer:

```
$ ~/git/jdk17u-dev/build/linux-x86_64-server-release/jdk/bin/javac Main.java && ~/git/jdk17u-dev/build/linux-x86_64-server-release/jdk/bin/java --version && ~/git/jdk17u-dev/build/linux-x86_64-server-release/jdk/bin/java Main
openjdk 17.0.11-internal 2024-04-16
OpenJDK Runtime Environment (build 17.0.11-internal+0-adhoc.delawen.jdk17u-dev)
OpenJDK 64-Bit Server VM (build 17.0.11-internal+0-adhoc.delawen.jdk17u-dev, mixed mode)
foo=should be foo, bar=3
foo=should be foo, bar=3
```

Delawen:
>  get some more testing than just checking that is fixes the issue.

@GoeLin I am going to test other things with my local build with the fix, like running other Java apps. Any suggestions on what should I try to make sure it works, besides the different `make tests` and try to run for example IntellijIDEA or other Java apps?

Delawen:
The commit contains a test for the different ways of calling a method with varargs: https://github.com/openjdk/jdk17u-dev/pull/2235/files#diff-41008904ed7aa596f5de5930e082a8cf6b239603883818fa940bb04ff511455aR65-R68

But I understand this may be a "too simple" example.

Delawen:
@mlchung You did the original commit for this fix. Can you give me some ideas on how to properly test this backport and make sure I'm not breaking anything else?

// cc @AlanBateman @DasBrain @liach

DasBrain:
I did not read through the entire patch - @mlchung did comment on the bug report back then:

> Alternative fix for backport consideration:
> 
>     @@ -343,7 +343,7 @@ public class MethodHandleProxies {
>                              intfc, mk.getName(),
>                              MethodType.methodType(mk.getReturnType(), mk.getParameterTypes()),
>                              self.getClass());
>     - return mh.asSpreader(Object[].class, mk.getParameterCount());
>     + return mh.asFixedArity().asSpreader(Object[].class, mk.getParameterCount()); 

https://bugs.openjdk.org/browse/JDK-8280377?focusedId=14471119&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14471119

mlchung:
The backport looks good to me.    This is the same fix for [JDK-8280377](https://bugs.openjdk.org/browse/JDK-8280377) and includes the regression test.   Tier1-3 tests including this new regression test should provide adequate verification.

liach:
The alternative backport fix is for JDK versions before 16 when Proxy.invokeDefault was not available, right?

mlchung:
> The alternative backport fix is for JDK versions before 16 when Proxy.invokeDefault was not available, right?

Yes.   `Proxy::invokeDefault` API was introduced in Java 16.

src/java.base/share/classes/java/lang/reflect/InvocationHandler.java
```diff
@@ -262,33 +261,6 @@ public static Object invokeDefault(Object proxy, Method method, Object... args)
             throws Throwable {
         Objects.requireNonNull(proxy);
         Objects.requireNonNull(method);
-
```
Delawen, line 265: This removed piece of code is moved to `Proxy.invokeDefault` (which is now called at the end of this removed code). The code moved is mostly the same, except it checks if the caller object is null. If it is  null, then it won't execute anything.

src/java.base/share/classes/java/lang/invoke/MethodHandleProxies.java
```diff
@@ -202,7 +202,8 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                     if (isObjectMethod(method))
                         return callObjectMethod(proxy, method, args);
                     if (isDefaultMethod(method)) {
-                        return callDefaultMethod(defaultMethodMap, proxy, intfc, method, args);
+                        // no additional access check is performed
+                        return JLRA.invokeDefault(proxy, method, args, null);
```
Delawen, line 206: As defined at the end of the file, `JavaLangReflectAccess JLRA = SharedSecrets.getJavaLangReflectAccess();`

src/java.base/share/classes/java/lang/reflect/Proxy.java
```diff
@@ -1314,6 +1314,46 @@ public MethodHandles.Lookup run() {
         });
     }
 
+    /*
+     * Invoke the default method of the given proxy with an explicit caller class.
+     *
+     * @throws IllegalAccessException if the proxy interface is inaccessible to the caller
+     *         if caller is non-null
+     */
+    static Object invokeDefault(Object proxy, Method method, Object[] args, Class<?> caller)
```
Delawen, line 1323: This is the piece of code that will now unify calling methods both from `MethodHandleProxies` (where this code is copied from) and `ReflectAccess`, which was using a different way of calling methods.

theRealAph (approved):
OK, I see now. Thanks for the explanation.

Delawen:
Although I miss the bot saying something

Delawen:
/approval JDK-8280377 request Fix Request
This fix is a backport. I am applying the same fix that was done for version 19. From the two ways of calling a method via interface with a variable number of arguments (array or varargs), it now uses the same way for both, with the code moved to Proxy.java class.
It includes tests to make sure that when we call MethodHandleProxies, it does correctly invoke default methods both with and without varags

Delawen:
Is that what you needed, bot?

GoeLin:
> The backport looks good to me. This is the same fix for [JDK-8280377](https://bugs.openjdk.org/browse/JDK-8280377) and includes the regression test. Tier1-3 tests including this new regression test should provide adequate verification.

Hi @Delawen, did you run some tests as e.g. proposed here?

Delawen:
> Hi @Delawen, did you run some tests as e.g. proposed here?

To be completely honest, I am having issues running tier2+ tests on my laptop, I am looking for an external infra to run them.

Delawen:
@smlambert is using this PR branch for polishing the Trestle testing pipeline. 

The results are here: https://github.com/adoptium/aqa-tests/issues/5137#issuecomment-1997362685

Not all architectures pass, but (at least right now) not because of this PR, but because misconfigurations.

// cc @GoeLin

Delawen:
Do we need more testing here?

jerboaa:
> Do we need more testing here?

I don't think we do. We [have positive tier 1-3 results](https://github.com/adoptium/aqa-tests/issues/5137#issuecomment-1997362685) and the regression test. It fixes a real bug. Patch is reviewed, even though it's clean.

Delawen:
/approval JDK-8280377 request Fix Request
This fix is a backport. I am applying the same fix that was done for version 19. From the two ways of calling a method via interface with a variable number of arguments (array or varargs), it now uses the same way for both, with the code moved to Proxy.java class.
It includes tests to make sure that when we call MethodHandleProxies, it does correctly invoke default methods both with and without varags

PR #2257, opened 2024-02-29
Backport 984003d5c969443abae2d889e92cba30da26e55f

snazarkin:
I'd like to backport this changes as it fixes some custom launchers issue.  Fix applies cleanly. 

The functionality was tested with the new test and manually
```
    cd ~/tools/bin
    cp -r ~/jdk .
    cp jdk/bin/java jdk/lib
    jdk/lib/java -version

```
Without the patch JVM can't start
```
Error: could not find libjava.so
Error: Could not find Java SE Runtime Environment.
```

### DIFF
--- a/src/java.base/unix/native/libjli/java_md_common.c
+++ b/src/java.base/unix/native/libjli/java_md_common.c
@@ -45,20 +45,27 @@ static char* findLastPathComponent(char *buffer, const char *comp) {
 /*
  * Removes the trailing file name and any intermediate platform
  * directories, if any, and its enclosing directory.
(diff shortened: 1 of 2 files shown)


GoeLin:
@snazarkin, why are the GHA tests failing?  Next time, I might just remove the fix-request label instead of asking. Testing is a prereq for labeling.

snazarkin:
> @snazarkin, why are the GHA tests failing? Next time, I might just remove the fix-request label instead of asking. Testing is a prereq for labeling.

It is riscv pipeline can't create sysroot. Looks like all PR for the last few days are failing

PR #2339, opened 2024-03-27
Backport 262747094670b00ac63463a059074afa9b81d8a4

elifaslan1:
Clean backport.Testing is in progress


elifaslan1:
The test does not work reliably, because JDK-8233725 bugfix is missing. I'll look into backporting that one first.

elifaslan1:
Here is the backport PR: https://github.com/openjdk/jdk17u-dev/pull/2456

elifaslan1:
/approval request Clean backport to address the segmentation fault issue that occurs when jspawnhelper is called without arguments. There is a new test added to verify the behavior in such cases.New regression test fails without the fix, and passes with the change.

PR #2635, opened 2024-06-25
8303466: C2: failed: malformed control flow. Limit type made precise with MaxL/MinL

martinuy:
Hi,

I would like to propose a backport of 8303466 [1] to jdk17u. jdk17u can benefit from this fix and having more accurate limit type information.

The JDK main line patch (introduced for JDK 21) does not apply cleanly because of the following:

 * src/hotspot/share/opto/convertnode.cpp
   * jdk17u does not have 8230382 so the context where the change applies is different. Manually applied the change.

 * src/hotspot/share/opto/loopTransform.cpp
   * jdk17u does not have 8294540 that removes the protection of the limit under an Opaque2Node. This is orthogonal to the change introduced by 8303466, so there is no need to make 8294540 a dependency.
   * jdk17u does not have 8299975 that assigns more accurate type information to the CMoveINode node. This is not necessary with 8303466 because the MaxLNode or MinLNode macro nodes (inserted instead of the CMoveINode node during PhaseIdealLoop) will have accurate type information, even more accurate than CMoveINode which used the original limit for one of the range ends (intead of the limit minus the strides after unrolling). See MaxLNode::add_ring and MinLNode::add_ring.
   * Note: even with 8299975 backported, 8303466 wouldn't apply cleanly because 8301074 was backported already and it applies changes to code introduced by 8299975 (these changes were not in 17u's 8301074 and would have to be applied manually).
  * The CastIINode node inserted during the first unrolling to keep precise type information after inserting an Opaque2 node is still needed, but has to be inserted before the node is used for ConvI2LNode -> SubLNode -> MaxLNode/MinLNode -> ... Otherwise, bottom type propagates accross this nodes path. The type of the node has to reflect the original limit type. MaxLNode/MinLNode will handle the clamping effect over the type.

In addition to the previous conflicts, the following changes were applied:

 * test/hotspot/jtreg/compiler/loopopts/TestLoopLimitSubtractionsCollapse.java
 * test/hotspot/jtreg/compiler/lib/ir_framework/IRNode.java
   * jdk17u does not have 8280378 (which, in turn, depends on 8280076 and is not in jdk17u either). These are large enhancements that extend the IR framework for tests and backporting them would be a large effort. The test is making an assertion in an intermediate graph to check the number of MaxL/MinL nodes, and make sure that the optimization that merges them (MaxLNode::Ideal/MinLNode::Ideal) was applied. These macro nodes are not longer available after the Macro expansion phase, when we can make assertions. I manually verified that the MaxL/MinL nodes optimization was applied both debugging fold_subI_no_underflow_pattern and analyzing the graph. If the optimization of MaxL/MinL nodes were not applied, there would be an impact on performance but should be harmless. I propose to skip this test.

No regressions observed in hotspot:tier1.

Thanks,
Martin.-

--
[1] - https://bugs.openjdk.org/browse/JDK-8303466

---------
### Progress
- [x] [JDK-8303466](https://bugs.openjdk.org/browse/JDK-8303466) needs maintainer approval

### Issue
 * [JDK-8303466](https://bugs.openjdk.org/browse/JDK-8303466): C2: failed: malformed control flow. Limit type made precise with MaxL/MinL (**Bug** - P3 - Approved)

### Reviewers
 * [Roland Westrelin](https://openjdk.org/census#roland) (@rwestrel - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/2635/head:pull/2635` \
`$ git checkout pull/2635`

`$ git checkout pull/2635` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/2635/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2635`

`$ git pr show -t 2635`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/2635.diff">https://git.openjdk.org/jdk17u-dev/pull/2635.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/2635#issuecomment-2187973185)

### DIFF
--- a/src/hotspot/share/opto/addnode.cpp
+++ b/src/hotspot/share/opto/addnode.cpp
@@ -1257,6 +1257,138 @@ const Type *MinINode::add_ring( const Type *t0, const Type *t1 ) const {
   return TypeInt::make( MIN2(r0->_lo,r1->_lo), MIN2(r0->_hi,r1->_hi), MAX2(r0->_widen,r1->_widen) );
 }
 
+// Collapse the "addition with overflow-protection" pattern, and the symmetrical
+// "subtraction with underflow-protection" pattern. These are created during the
+// unrolling, when we have to adjust the limit by subtracting the stride, but want
+// to protect against underflow: MaxL(SubL(limit, stride), min_jint).
+// If we have more than one of those in a sequence:
+//
+//   x  con2
+//   |  |
+//   AddL  clamp2
+//     |    |
+//    Max/MinL con1
+//          |  |
+//          AddL  clamp1
+//            |    |
+//           Max/MinL (n)
+//
+// We want to collapse it to:
+//
+//   x  con1  con2
+//   |    |    |
+//   |   AddLNode (new_con)
+//   |    |
+//  AddLNode  clamp1
+//        |    |
+//       Max/MinL (n)
+//
+// Note: we assume that SubL was already replaced by an AddL, and that the stride
+// has its sign flipped: SubL(limit, stride) -> AddL(limit, -stride).
+Node* fold_subI_no_underflow_pattern(Node* n, PhaseGVN* phase) {
+  assert(n->Opcode() == Op_MaxL || n->Opcode() == Op_MinL, "sanity");
+  // Check that the two clamps have the correct values.
+  jlong clamp = (n->Opcode() == Op_MaxL) ? min_jint : max_jint;
+  auto is_clamp = [&](Node* c) {
+    const TypeLong* t = phase->type(c)->isa_long();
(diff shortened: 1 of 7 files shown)


rwestrel (approved):
That looks good to me.

martinuy:
The GHA macos-aarch64 failure at build time seems unrelated to this backport.

martinuy:
/approval request jdk17u is affected by this bug and would benefit from its fix. The patch does not apply cleanly but has been reviewed.

PR #2688, opened 2024-07-06
8335808: update for deprecated sprintf for jfrTypeSetUtils

gdams:
The occurrence of `sprintf` was removed in JDK18 as part of https://bugs.openjdk.org/browse/JDK-8266936. This patch simply converts it to `snprintf` to avoid build failures.


MBaesken (approved):

gdams:
/approval request Low risk, all tests pass. This is also crucial to allow us to upgrade the GH actions workflow to move away from the deprecated macos-11 VMs.

PR #2829, opened 2024-08-28
8332825: ubsan: guardedMemory.cpp:35:11: runtime error: null pointer passed as argument 2, which is declared to never be null

MBaesken:
---------
### Progress
- [x] [JDK-8332825](https://bugs.openjdk.org/browse/JDK-8332825) needs maintainer approval

### Issue
 * [JDK-8332825](https://bugs.openjdk.org/browse/JDK-8332825): ubsan: guardedMemory.cpp:35:11: runtime error: null pointer passed as argument 2, which is declared to never be null (**Bug** - P4 - Approved)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/2829/head:pull/2829` \
`$ git checkout pull/2829`

`$ git checkout pull/2829` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/2829/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2829`

`$ git pr show -t 2829`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/2829.diff">https://git.openjdk.org/jdk17u-dev/pull/2829.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/2829#issuecomment-2315231888)

### DIFF
--- a/src/hotspot/share/memory/guardedMemory.cpp
+++ b/src/hotspot/share/memory/guardedMemory.cpp
@@ -33,7 +33,9 @@ void* GuardedMemory::wrap_copy(const void* ptr, const size_t len, const void* ta
   if (outerp != NULL) {
     GuardedMemory guarded(outerp, len, tag);
     void* innerp = guarded.get_user_ptr();
-    memcpy(innerp, ptr, len);
+    if (ptr != nullptr) {
+      memcpy(innerp, ptr, len);
+    }
     return innerp;
   }
   return NULL; // OOM


src/hotspot/share/memory/guardedMemory.cpp
```diff
@@ -33,7 +33,9 @@ void* GuardedMemory::wrap_copy(const void* ptr, const size_t len, const void* ta
   if (outerp != NULL) {
     GuardedMemory guarded(outerp, len, tag);
     void* innerp = guarded.get_user_ptr();
-    memcpy(innerp, ptr, len);
+    if (ptr != nullptr) {
+      memcpy(innerp, ptr, len);
+    }
```
tstuefe, line 38: Hmm. innerp can never be null. If anything, we should assert.

GoeLin: Hi,
I don't understand. First, innerp is not checked for null. Second: does the code in 17 differ to 21&head? Else we should change this to an assertion in head, first.

GoeLin:
GHA failure: ClhsdbFindPC is known to fail in GHA on mac aarch. Here we see a different error situation, the test times out. But this surely is unrelated to this fix.

src/hotspot/share/memory/guardedMemory.cpp
tstuefe: Hi Goetz!

`innerp` is the result of 

```
  u_char* get_user_ptr() const {
    assert(_base_addr != nullptr, "Not wrapping any memory");
    return _base_addr + sizeof(GuardHeader);
  }
```

which cannot return null. _base_addr is the result of a malloc and we only ever enter this path if that malloc succeeded. It is also asserted at least twice.

Cheers, Thomas

GoeLin: Yes, but the new code checks whether "ptr" is null. That is passed in as an argument.

tstuefe: Okay, I just misread the code. Sorry for the confusion, I was too tired.

PR #3038, opened 2024-11-11
Backport jdk 8341779

cushon:


fitzsim:
Thank you for providing this draft.  I replicated the crash in my environment, confirmed the lines added to the test are required to trigger the crash, and confirmed that reverting "Disable handling of annotations with different retention at the same location" fixes the crash.

PR #3078, opened 2024-11-29
Backport f3be138eb80c9e7f6cc21afb75cda9e49b667c8a

GoeLin:
This is a backport of the 21 PR code, but it needs further adaptions.

SystemModulesPlugin.java had to be resolved. The code to be patched does the 
same, but uses different Java syntax. 

Also, I had to remove one more case from the test as jdk.httpserver has no main
specified.
But the test still shows the same failure as in head when the fix misses
in the JVM, and passes if the fix is present.


GoeLin:
In the second commit, I removed the modules not suitable for this test as I did in 21.

MBaesken (approved):

PR #3278, opened 2025-02-17
8302111: Serialization considerations

GoeLin:
I backport this to match 17.0.13-oracle based on the commit to 21.

I had to resolve several files, two of them considerably:

src/java.base/share/classes/com/sun/crypto/provider/DHPrivateKey.java
  Resolved larger chunk.
src/java.base/share/classes/com/sun/crypto/provider/DHPublicKey.java
  Resolved some code.

The new code in DHPublic/PrivateKey depends on the removal
of IOExceptions in [JDK-8297065](https://bugs.openjdk.org/browse/JDK-8297065) "DerOutputStream operations should not throw IOExceptions". 
This change is in 21 but not in 17. Thus code backported from 21 calls functions of 
DerOutputStream that do not throw an exception in 21, but do so in 17. 
JDK-8297065 makes the point that these exceptions can never be thrown, which also
holds for 17.  So I just catch and ignore them.

Also, I had to adapt code because an exception constructor with cause for
InvalidObjectException is missing in 17.
This was added by [JDK-8282696](https://bugs.openjdk.org/browse/JDK-8282696) "Add constructors taking a cause
to InvalidObjectException and InvalidClassException" in 19.
I added a call to initCause() to store the causing exception.

I double-checked that [JDK-8297065](https://bugs.openjdk.org/browse/JDK-8297065) was not backported
by Oracle. Backporting this change would simplify matters considerably.

The remaining files only needed trivial resolves:

src/java.base/share/classes/java/security/Permissions.java
src/java.base/share/classes/java/security/SignedObject.java
  Only Copyright.
src/java.base/share/classes/java/security/Timestamp.java
  Copyright and import.
src/java.base/share/classes/java/security/UnresolvedPermissionCollection.java
src/java.base/share/classes/java/security/cert/CertificateRevokedException.java
src/java.base/share/classes/sun/security/provider/DRBG.java
src/java.base/share/classes/sun/security/util/ObjectIdentifier.java
src/java.base/share/classes/sun/security/x509/AlgIdDSA.java
  Only Copyright.
src/java.security.jgss/share/classes/sun/security/jgss/krb5/Krb5Context.java
  Resolved imports.
src/java.security.jgss/share/classes/sun/security/jgss/krb5/Krb5InitCredential.java
src/jdk.crypto.cryptoki/share/classes/sun/security/pkcs11/P11SecureRandom.java
  Only Copyright.

I manually ran these tests on linux x86_64:

test/jdk/com/sun/security
test/jdk/java/awt/security
test/jdk/java/net/httpclient/security
test/jdk/java/net/httpclient/websocket/security
test/jdk/java/security
test/jdk/javax/management/security
test/jdk/javax/security
test/jdk/jdk/security
test/jdk/security
test/jdk/sun/security

Our well-known nightly tests passed, too.

---------
### Progress
- [x] [JDK-8302111](https://bugs.openjdk.org/browse/JDK-8302111) needs maintainer approval

### Issue
 * [JDK-8302111](https://bugs.openjdk.org/browse/JDK-8302111): Serialization considerations (**Bug** - P4 - Approved)

### Reviewers
 * [Richard Reingruber](https://openjdk.org/census#rrich) (@reinrich - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/3278/head:pull/3278` \
`$ git checkout pull/3278`

`$ git checkout pull/3278` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/3278/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 3278`

`$ git pr show -t 3278`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/3278.diff">https://git.openjdk.org/jdk17u-dev/pull/3278.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/3278#issuecomment-2663626233)
</details>

### DIFF
--- a/src/java.base/share/classes/com/sun/crypto/provider/DHPrivateKey.java
+++ b/src/java.base/share/classes/com/sun/crypto/provider/DHPrivateKey.java
@@ -162,7 +162,8 @@ final class DHPrivateKey implements PrivateKey,
     private static byte[] encode(BigInteger p, BigInteger g, int l,
             byte[] key) {
         DerOutputStream tmp = new DerOutputStream();
-
+        byte[] encoded;
+        try {
         // version
         tmp.putInteger(PKCS8_VERSION);
 
@@ -191,10 +192,13 @@ final class DHPrivateKey implements PrivateKey,
 
         // make it a SEQUENCE
         DerValue val = DerValue.wrap(DerValue.tag_Sequence, tmp);
-        byte[] encoded = val.toByteArray();
+        encoded = val.toByteArray();
         val.clear();
-
-        return encoded;
+        } catch (IOException e) {
+            // Ignore, see JDK-8297065.
+            encoded = null;
+        }
+      return encoded;
     }
 
     /**
@@ -234,6 +238,8 @@ final class DHPrivateKey implements PrivateKey,
         DerValue val = new DerValue(DerValue.tag_Integer, xbytes);
         try {
             this.key = val.toByteArray();
+        } catch (IOException e) {
+            // Ignore, see JDK-8297065.
         } finally {
             val.clear();
             Arrays.fill(xbytes, (byte) 0);
@@ -369,7 +375,11 @@ final class DHPrivateKey implements PrivateKey,
         try {
             c = decode(encodedKeyIntern);
         } catch (IOException e) {
-            throw new InvalidObjectException("Invalid encoding", e);
+            InvalidObjectException ioe = new InvalidObjectException("Invalid encoding");
+            if (ioe != null) {
+                ioe.initCause(e);
+            }
+            throw ioe;
         }
         if (!Arrays.equals(c.key, key) || !c.x.equals(x) || !c.p.equals(p)
                 || !c.g.equals(g) || c.l != l) {
--- a/src/java.base/share/classes/com/sun/crypto/provider/DHPrivateKey.java
+++ b/src/java.base/share/classes/com/sun/crypto/provider/DHPrivateKey.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -57,6 +57,7 @@ final class DHPrivateKey implements PrivateKey,
     private BigInteger x;
 
     // the key bytes, without the algorithm information
+    // cannot be final as it's re-assigned for deserialization
     private byte[] key;
 
     // the encoded key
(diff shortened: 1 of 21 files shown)


src/java.base/share/classes/com/sun/crypto/provider/DHPublicKey.java
```diff
@@ -71,6 +72,122 @@ final class DHPublicKey implements PublicKey,
...
+        tmpDerKey.putBitString(key);
+
+        // wrap algid and key into SEQUENCE
+        derKey = new DerOutputStream();
+        derKey.write(DerValue.tag_Sequence, tmpDerKey);
+        } catch (IOException e) {
+            // Ignore, see JDK-8297065.
+            derKey = null;
+        }
+        return derKey.toByteArray();
```
reinrich, line 188: This looks weired if someone has a quick look at these lines of code: if an IOE occurs and is caught here, then `derKey` will be assigned null and 2 lines below an NPE will thrown because of this.

I'd suggest to wrap the whole method body in a try-catch. This would reduce the diff to jdk 21.
In the IOE catch clause you shouldn't say that the IOE is ignored but you should state that it cannot even occur since `DerOutputStream` is a `ByteArrayOutputStream` which doesn't do any I/O. Then just return null or throw an `InternalError` are something else if more appropriate.

src/java.base/share/classes/com/sun/crypto/provider/DHPublicKey.java
```diff
@@ -106,7 +223,7 @@ final class DHPublicKey implements PublicKey,
         try {
             this.key = new DerValue(DerValue.tag_Integer,
                                     this.y.toByteArray()).toByteArray();
-            this.encodedKey = getEncoded();
+            this.encodedKey = encode(p, g, l, key);
         } catch (IOException e) {
             throw new ProviderException("Cannot produce ASN.1 encoding", e);
         }
```
reinrich, line 229: You should remove the try-catch. `encode` doesn't throw IOE. This also removes the diff to jdk 21.

src/java.base/share/classes/com/sun/crypto/provider/DHPublicKey.java
```diff
@@ -339,11 +368,32 @@ private void readObject(ObjectInputStream stream)
...
+        // check if the "encodedKey" value matches the deserialized fields
+        DHComponents c;
+        byte[] encodedKeyIntern = encodedKey.clone();
+        try {
+            c = decode(encodedKeyIntern);
+        } catch (IOException e) {
+            InvalidObjectException ioe = new InvalidObjectException("Invalid encoding");
+            if (ioe != null) {
+                ioe.initCause(e);
+            }
```
reinrich, line 384: The null check is redundant.
```suggestion
            ioe.initCause(e);
```

reinrich (changes requested):

src/java.base/share/classes/com/sun/crypto/provider/DHPrivateKey.java
```diff
@@ -351,11 +365,32 @@ private void readObject(ObjectInputStream stream)
...
         }
-        this.encodedKey = encodedKey.clone();
+        // check if the "encodedKey" value matches the deserialized fields
+        DHComponents c;
+        byte[] encodedKeyIntern = encodedKey.clone();
+        try {
+            c = decode(encodedKeyIntern);
+        } catch (IOException e) {
+            InvalidObjectException ioe = new InvalidObjectException("Invalid encoding");
+            if (ioe != null) {
```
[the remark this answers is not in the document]
GoeLin: My inital desing was 
throw new InvalidObjectException("Invalid encoding").initCause(e);
But the compiler does not like this:
error: unreported exception Throwable; must be caught or declared to be thrown
So I decided to add the try/catch with the null check.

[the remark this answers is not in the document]
reinrich: I see... strange. But it does compile without the null check. I think it should be removed.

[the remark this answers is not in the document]
GoeLin: I'll remove the null check in both places where I added initCause().

src/java.base/share/classes/com/sun/crypto/provider/DHPublicKey.java
GoeLin: It is toByteArray that trows the IOException.
I can add the InternalError here as in the other places?

src/java.base/share/classes/com/sun/crypto/provider/DHPublicKey.java
GoeLin: Removed.

GoeLin:
Triggered by the discussion here and offline I had a second look at the exceptions I need to add because JDK-8297065 is not in 17.  I committed some code that brings these closer to the origin. One I had added is not really needed as IOException is caught at the callsite anyways.

reinrich:
> Triggered by the discussion here and offline I had a second look at the exceptions I need to add because JDK-8297065 is not in 17. I committed some code that brings these closer to the origin. One I had added is not really needed as IOException is caught at the callsite anyways.

And the diff to the versions of DHPublicKey.java and DHPrivateKey.java is minimal now.

GoeLin:
> > Triggered by the discussion here and offline I had a second look at the exceptions I need to add because JDK-8297065 is not in 17. I committed some code that brings these closer to the origin. One I had added is not really needed as IOException is caught at the callsite anyways.
> 
> And the diff to the versions of DHPublicKey.java and DHPrivateKey.java is minimal now.

Yes, thanks for reviewing!

GoeLin:
Hi @martinuy,
Although I think this is good to go now, I would appreciate a second look by you. Could you find time in the next days?  
Thanks, Goetz.

PR #3328, opened 2025-03-05
8298061: vmTestbase/nsk/sysdict/vm/stress/btree/btree012/btree012.java failed with "fatal error: refcount has gone to zero"

dchuyko:
Please review the backport of the fix for JDK-8298061 to jdk17u. The bug leads to rare crashes. For instance, they cause the Spring Boot build to fail under certain circumstances. In particular on AMD EPYC 7R13 and Ubuntu 24.04.1.

Original patch doesn't apply cleanly. First, technically clear_supername() was only added by JDK-8292286 in JDK 20 and Symbol::maybe_increment/decrement_refcount were added by JDK-8291457 also in JDK 20, all before the original patch. 
Second, PlaceholderTable::find_and_remove() actually needs what clear_supername() call does in 20 to make the fix complete (and to match find_and_add()). That second part was later reworked in JDK 21 (JDK-8302108) and after.

PlaceholderEntry::set_supername() was modified to perform same work as before under the same condition as in the original patch.

The early version of the original change by @coleenp had an implementation of the second part which is suitable for 17u so it was made a part of this backport. See https://github.com/openjdk/jdk/pull/11726

Without that second part part we did observe assert failures.

Testing: jtreg tier 1-3 on linux-amd64, dedicated stress test of the Spring Boot build.

---------
### Progress
- [x] [JDK-8298061](https://bugs.openjdk.org/browse/JDK-8298061) needs maintainer approval

### Issue
 * [JDK-8298061](https://bugs.openjdk.org/browse/JDK-8298061): vmTestbase/nsk/sysdict/vm/stress/btree/btree012/btree012.java failed with "fatal error: refcount has gone to zero" (**Bug** - P3 - Approved)

### Reviewers
 * [Coleen Phillimore](https://openjdk.org/census#coleenp) (@coleenp - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/3328/head:pull/3328` \
`$ git checkout pull/3328`

`$ git checkout pull/3328` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/3328/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 3328`

`$ git pr show -t 3328`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/3328.diff">https://git.openjdk.org/jdk17u-dev/pull/3328.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/3328#issuecomment-2701838743)
</details>

### DIFF
--- a/src/hotspot/share/classfile/placeholders.hpp
+++ b/src/hotspot/share/classfile/placeholders.hpp
@@ -142,8 +142,12 @@ class PlaceholderEntry : public HashtableEntry<Symbol*, mtClass> {
 
   Symbol*            supername()           const { return _supername; }
   void               set_supername(Symbol* supername) {
-    _supername = supername;
-    if (_supername != NULL) _supername->increment_refcount();
+    if (supername != _supername) {
+      _supername = supername;
+      if (_supername != NULL) {
+        _supername->increment_refcount();
+      }
+    }
   }
 
   Thread*            definer()             const {return _definer; }
--- a/src/hotspot/share/classfile/placeholders.cpp
+++ b/src/hotspot/share/classfile/placeholders.cpp
@@ -326,7 +326,12 @@ void PlaceholderTable::find_and_remove(unsigned int hash,
     PlaceholderEntry *probe = get_entry(hash, name, loader_data);
     if (probe != NULL) {
        log(probe, "find_and_remove", action);
-       probe->remove_seen_thread(thread, action);
+
+       bool empty = probe->remove_seen_thread(thread, action);
+       if (empty && action == LOAD_SUPER) {
+          probe->set_supername(NULL);
+       }
+
        // If no other threads using this entry, and this thread is not using this entry for other states
        if ((probe->superThreadQ() == NULL) && (probe->loadInstanceThreadQ() == NULL)
           && (probe->defineThreadQ() == NULL) && (probe->definer() == NULL)) {


src/hotspot/share/classfile/placeholders.hpp
```diff
@@ -142,8 +142,12 @@ class PlaceholderEntry : public HashtableEntry<Symbol*, mtClass> {
 
   Symbol*            supername()           const { return _supername; }
   void               set_supername(Symbol* supername) {
-    _supername = supername;
-    if (_supername != NULL) _supername->increment_refcount();
+    if (supername != _supername) {
```
coleenp, line 145: Why isn't there an if (_supername == NULL) _supername->decrement_refcount(); before assigning the new supername?

dchuyko: Also it turn out that checking for NULL may be not enough. PlaceholderEntry has no constructor, so members can contain garbage.

coleenp: I see.  This is the old hashtable implementation.

coleenp (approved):
This looks good. Thank you for testing out this backport.

src/hotspot/share/classfile/placeholders.hpp
dchuyko: Coleen, thanks for the review.

dchuyko:
/approval request This is a backport of a small reference counting fix in PlaceholderEntry::set_supername(). The bug leads to rare crashes. In particular, it causes the Spring Boot build to fail under certain circumstances. The original patch doesn't apply cleanly and is not enough (because of changes in 20 and later), it is slightly modified/extended. We basically add guards t prevent null/uninitialized data access so the risk is low. 

Testing: jtreg tier 1-3 on linux-amd64, dedicated stress test of the Spring Boot build.

PR #3330, opened 2025-03-07
Backport b6935dfb86a1c011355d2dfb2140be26ec536351

satyenme:
Backporting JDK-8251505: Use of types in compiler shared code should be consistent. Cleans up mixed types and usage of 'long' type. Ran GHA Sanity Checks, local Tier 1 and Tier 2 tests, and jdk/jfr and java/lang/management explicitly. Patch is clean.


satyenme:
/approval request for backport. Minor change that cleans up mixed types and usage of 'long' type. Ran GHA Sanity Checks, and local Tier 1 and Tier 2 tests. Risk is low.

GoeLin:
Please edit your fix request comment and tell me why you want to backport this and why you think the risk is low. Thanks for the testing. 
You can just issue a second approval command, it replaces the existing comment in JBS as far as I know.

PR #3409, opened 2025-03-27
Backport 9f98136c3a00ca24d59ffefd58308603b58110c7

cost0much:
temp


cost0much:
/approval request Fixes bug with `JarURLConnection` leaking file handles when `FileURLConnection` is used. Applies cleanly. Patched in JDK22, and no known bugs from this change. New regression test fails without the fix, and passes with it. Risk is low given the length of time in mainline and small change.

PR #3594, opened 2025-05-19
Backport 004aaea76db091569aa88eeb6b08db3408f288cd

satyenme:
Backporting JDK-8342330: C2: "node pinned on loop exit test?" assert failure. Adjusts condition to avoid assert failure edge case. Ran GHA Sanity Checks, local Tier 1 and 2, and new test. Patch is clean.


satyenme:
/approval request for backport. Adjusts condition to avoid assert failure edge case.

Risk: Medium-Low. Tip change was merged in October 2024, slight adjustment to logic to avoid assertion failure, and adds test to confirm.

Testing:
* GHA Sanity Checks
* Tier 1 and 2 tests locally
* New test

PR #3607, opened 2025-05-22
Backport 37ec796255ae857588a5c7e0d572407dd81cbec9

tstuefe:
This is not a clean backport. The effected G1Allocator and G1Collector methods have changed since JDK17.

So this backports reimplements the patch in a minimally invasive way while retaining as much similarity as possible with the original patch.

The gist of the patch is clear: instead of finding out the NUMA node index at every instance of G1Allocator::allocate_xxx, and then be subject to NUMA node migrations, we fix the NUMA node index once and use that one.

I tested this patch with my "FakeNUMA" addition (I plan to upstream that one at some point). This FakeNUMA mode mimics a lot of NUMA node migrations. I can verify that without this patch the JVM crashes quickly, with the patch it does not crash.

### DIFF
--- a/src/hotspot/share/gc/g1/g1Allocator.hpp
+++ b/src/hotspot/share/gc/g1/g1Allocator.hpp
@@ -88,9 +88,6 @@ private:
                                    size_t desired_word_size,
                                    size_t* actual_word_size);
 
-  // Node index of current thread.
-  inline uint current_node_index() const;
-
 public:
   G1Allocator(G1CollectedHeap* heap);
   ~G1Allocator();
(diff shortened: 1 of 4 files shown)


tstuefe:
Ping @sjohanss - could I get a review for this?

tstuefe:
@mdoerr or @MBaesken : would it be possible to send this through your JDK17 CI too, in case our tests missed something? Thanks!

MBaesken:
Yes we can test it.

GoeLin:
Hi,@tstuefe,  I'll remove the fix labels.  Could you please also mention in your fix-request comment that you withdrew it and why?  For the next one who might try backporting this?

PR #3664, opened 2025-06-17
Backport 2b57f402c46104d4aba784a891ba90604f5e9e4c

satyenme:
Backporting JDK-8343426: ConcurrentSkipListMap.spliterator() can no longer split the stream. Since Java 10, the trySplit() method for spliterators for the ConcurrentSkipListMap always returned null, causing parallel streams to be broken for ConcurrentSkipListMap and ConcurrentSkipListSet since. This change fixes the logic for considering the head.node.next field in addition to head.node. Ran GHA Sanity Checks, local Tier 1 and 2, and new test directly. Patch is clean.

### DIFF
--- a/src/java.base/share/classes/java/util/concurrent/ConcurrentSkipListMap.java
+++ b/src/java.base/share/classes/java/util/concurrent/ConcurrentSkipListMap.java
@@ -3193,14 +3193,14 @@ public class ConcurrentSkipListMap<K,V> extends AbstractMap<K,V>
     }
     // factory method for KeySpliterator
     final KeySpliterator<K,V> keySpliterator() {
-        Index<K,V> h; Node<K,V> n; long est;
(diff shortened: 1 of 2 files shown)


satyenme:
/approval request for backport of JDK-8343426: ConcurrentSkipListMap.spliterator() can no longer split the stream

Motivation: Without this change the trySplit() method for spliterators for ConcurrentSkipListMaps and ConcurrentSkipListSets using parallel streams will remain broken. With the change, the pre-Java 10 functionality will be restored, fixing the current buggy implementation. It's possible someone's written code that depends on the buggy implementation (which always returns null), but quite unlikely.

Risk: Low. Ran GHA Sanity Checks, local Tier 1 and 2, and new test directly. Patch is clean. Change has been present in tip since November, 2024.

satyenme:
Closing due to `jdk21u-fix-no` label

PR #3671, opened 2025-06-24
Backport ad7a8e86e0334390f87ae44cf749d2b47f1409a1

cost0much:
temp


cost0much:
/approval request Fixes CompileTask timestamp int overflow when > 24.8 days. Applies cleanly from JDK22 and no bugtail. Passes GHA. Risk is low-medium as only log change that doesn't impact runtime but does touch hotspot.

PR #3672, opened 2025-06-24
Backport c47a0ce3f03daf7a4ec0c0c8b8626bb662836936

cost0much:
temp


cost0much:
/approval request Cleans up incorrect Access API usage in two nmethod methods. Applies cleanly from JDK22 and no bugtail. Passes GHA. Risk is low-medium as minor change that shouldn't impact runtime but does touch hotspot code.

PR #3809, opened 2025-08-04
backport 80edd5c298f21c5e5be3a0c2bb63129e76e0334f

GoeLin:
I would like to fix this in 17, too.  
Nice to have fix, good repro case, and makes backorting JDK-8349583 more easy.


GoeLin:
GHA failure: unrelated.
corss compile failures fixed by JDK-8363965: GHA: Switch cross-compiling sysroots to Debian bookworm

PR #3786, opened 2025-07-29
8342075: HttpClient: improve HTTP/2 flow control checks

GoeLin:
I backport this for parity with 17.0.17-oracle based on the change in 21.

src/java.net.http/share/classes/jdk/internal/net/http/Stream.java
Two trivial resolves due to context.

src/java.net.http/share/classes/jdk/internal/net/http/WindowUpdateSender.java

Resolved because 8308310: HttpClient: Avoid logging or locking from within synchronized blocks
is not in 17.

src/java.net.http/share/classes/module-info.java
Omitted, as "8296804: Document HttpClient configuration properties in java.net.http module-info"
is not in 17.  The CSR for 17 documents that this is dropped.

Follow ups are on the list of backports I will do.
Blocking changes all already in 17.

I had to adapt ConnectionFlowControlTest.java and StreamFlowControlTest.java because HttpClient does not
implement AutoClosable / shutdownNow() / close() in 17.
I also amended the @build instructions in StreamFlowControlTest.java.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] [JDK-8342075](https://bugs.openjdk.org/browse/JDK-8342075) needs maintainer approval
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change requires CSR request [JDK-8351085](https://bugs.openjdk.org/browse/JDK-8351085) to be approved

### Issues
 * [JDK-8342075](https://bugs.openjdk.org/browse/JDK-8342075): HttpClient: improve HTTP/2 flow control checks (**Bug** - P4 - Approved)
 * [JDK-8351085](https://bugs.openjdk.org/browse/JDK-8351085): HttpClient: improve HTTP/2 flow control checks (**CSR**)

### Reviewers
 * [Alexey Bakhtin](https://openjdk.org/census#abakhtin) (@alexeybakhtin - **Reviewer**) Review applies to [211a4553](https://git.openjdk.org/jdk17u-dev/pull/3786/files/211a4553461921ef6134ef05b300cc8128ae9274)
 * [Ralf Schmelter](https://openjdk.org/census#rschmelter) (@schmelter-sap - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/3786/head:pull/3786` \
`$ git checkout pull/3786`

[...]

PR #3852, opened 2025-08-14
Backport bd6152f5967107d7b32db9bcfa224fc07314f098

GoeLin:
I backport this for parity with 17.0.17-oracle.
Also, it is a follow-up to JDK-8342075.

It needed resolves and adaptions:

**Resolved** the first chunk in Stream.java which adds the declaration of inputQLock.
The declaraions in the context have been modified by "8308310: HttpClient: Avoid logging or locking from within synchronized blocks".
Omitted one chunk fixing white space.
See first commit.

**Adaptions** can be found in the second commit.

The change uses locks from 8308310 and introduces a new lock.
To make this work further adaptions are needed (see extra commit):
I add edimports required by the new lock inputQLock.

stateLock is used by this change, but only introduced by 8308310.
Where 21 uses stateLock, the implementation in 17 locks
itself ("this").  I changed the use of stateLock introduced
by this change in pushDateaFrame() accordingly.

Actually, 17 declares a row of methods as synchronized
that are not handled by stateLock in 21, e.g. isCanceled().
So there is a difference in behavior wrt. this lock
between 17 and 21. 

I needed to fix two build issues in StreamFlowControlTest.java:

.../test/jdk/java/net/httpclient/http2/StreamFlowControlTest.java:299: error: incompatible types: try-with-resources not applicable to variable type
        try (var client = HttpClient.newBuilder().sslContext(sslContext).build()) {
                 ^
    (HttpClient cannot be converted to AutoCloseable)

/.../test/jdk/java/net/httpclient/http2/StreamFlowControlTest.java:300: error: cannot find symbol
            var request = HttpRequest.newBuilder(URI.create(h2Head)).HEAD().build();
                                                                    ^
  symbol:   method HEAD()
  location: interface Builder

The first I resolved similar to other places in this test in the backport that
brought the test to 17:
https://github.com/openjdk/jdk17u-dev/pull/3786/commits/073bb3d5b0391cf9ad419a9b2be0d30afba1427b
Thanks to Aleksey for this detail.

For the second issue I replaced HEAD() by it's implementation.


phohensee (approved):

schmelter-sap (approved):
Looks good.

GoeLin:
Thanks for the reviews!

PR #4402, opened 2026-05-04
8383659: [17u] JVM crashes during stub routines generation on Windows and rare combination of CPU features

fthevenet:
When running on Windows with a very specific set of CPU flags available to the JVM, it crashes when generating stub routines because the allocated code buffer is too small.

The crash occurs specifically when running a CPU that does *not* advertise support the dedicated `sha` instruction set, but does claim support for all other modern CPU futures, notably`avx512` instructions.
Whereas in the absence of the `sha` features some stubs related to sha1 operations are *not* generated, the JVM however falls back to an avx2-based implementation for sha256 operations, which results in much larger blobs of generated assembly code compared to the fast_sha256 impl:

```
Using sha256_AVX2:
sha256_implCompress [0x00007fb3d86a0120, 0x00007fb3d86a124f] 4399 bytes
sha256_implCompressMB [0x00007fb3d86a1260, 0x00007fb3d86a23bd] 4445 bytes

Using fast_sha256:
sha256_implCompress [0x00007fbd14437ea0, 0x00007fbd144381cd] 813 bytes
sha256_implCompressMB [0x00007fbd144381e0, 0x00007fbd14438521] 833 bytes
```

This is enough to push the size of code generated by `StubRoutines::initialize2` above the 60300 bytes allocated to that specific buffer and cause a crash.
Please note that the issue has only been observe on Windows, due to stubs being larger due to abi differences, compared to Linux.

Disabling support for avx512 with `-XX:UseAVX=2` is an effective workaround, because the stubs for avx512 contributes greatly to the code buffer's occupation.

The crash is systematic (i.e. `java --version` is enough to provoke it) but the set of circumstances for it, is quite rare, as recent x86_64 CPU from both AMD and Intel do support the `sha` instruction set, and older models that don't, generally do not support avx512.

There is however, one specific case where this has been observed with consistency which is customers when running their jvm on virtual machines, specifically KVM+QEMU, where the CPU model is overridden to "Icelake-Server" (or any cpu model that matches version 4 of Qemu's ABI compatility level[0], where `sha` is conspicuously absent).
In this case, it is generally very difficult or impossible for the teams whose applications are affected to obtain a change in configuration from their infrastructure provider.

[0] https://www.qemu.org/docs/master/system/qemu-cpu-models.html

---------
- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

---------
### Progress
- [x] [JDK-8383659](https://bugs.openjdk.org/browse/JDK-8383659) needs maintainer approval

### Issue
 * [JDK-8383659](https://bugs.openjdk.org/browse/JDK-8383659): [17u] JVM crashes during stub routines generation on Windows and rare combination of CPU features (**Bug** - P4 - Approved)

### Reviewers
 * [Thomas Fitzsimmons](https://openjdk.org/census#fitzsim) (@fitzsim - no project role) Review applies to [b733387a](https://git.openjdk.org/jdk17u-dev/pull/4402/files/b733387ab40c65a7b39cc7166033ce48cb0341b9)
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**) Review applies to [b733387a](https://git.openjdk.org/jdk17u-dev/pull/4402/files/b733387ab40c65a7b39cc7166033ce48cb0341b9)
 * [Andrew John Hughes](https://openjdk.org/census#andrew) (@gnu-andrew - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/4402/head:pull/4402` \
`$ git checkout pull/4402`

`$ git checkout pull/4402` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/4402/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 4402`

`$ git pr show -t 4402`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/4402.diff">https://git.openjdk.org/jdk17u-dev/pull/4402.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/4402#issuecomment-4369445363)
</details>


fthevenet:
Please note that this particular issue does not affect jdk21 and later; it was addressed by increasing the size of the code buffer when running on Windows, in the same way this patch does, but this was done as a side effect of [JDK-828416: Implementation of Virtual Threads (Preview)](https://bugs.openjdk.org/browse/JDK-8284161), meaning a backport is quite obviously out of a scope here.

fitzsim (approved):
The `3200` value worked for me on a Windows virtual machine running on a hypervisor with an Intel CPU.  The virtual machine was configured with the settings you provided such that the virtual CPU had the exact feature set that triggered the crash.

fthevenet:
Thank you for testing this @fitzsim

fthevenet:
The pre-submit tests for building on Windows fail for both aarch64 and x64, because the CI job cannot install the the MSVC toolchain on the build worker.
I have no idea why this fails; I tried restarting the failed jobs a couple of times since yesterday, hoping this might be a transient problem, but got the same result.
This is pretty odd since the first commit in this PR (https://github.com/openjdk/jdk17u-dev/pull/4402/commits/481549f7523066f9aad894cc7d544bd29c66707e) had no such issue.

EDIT: I since restarted the pre-submit test jobs on that commit which originally passed, and they now fail too, further suggesting that something changes in the gha infrastructure rather than a fault in the code.

EDIT 2: This confirms this is an github infra related issue: https://github.com/actions/runner-images/issues/14004

src/hotspot/cpu/x86/stubRoutines_x86.hpp
```diff
@@ -32,8 +32,10 @@
 static bool returns_to_call_stub(address return_pc) { return return_pc == _call_stub_return_address; }
 
 enum platform_dependent_constants {
-  code_size1 = 20000 LP64_ONLY(+10000),         // simply increase if too small (assembler will crash if too small)
-  code_size2 = 35300 LP64_ONLY(+25000)          // simply increase if too small (assembler will crash if too small)
+  code_size1 = 20000 LP64_ONLY(+10000),                     // simply increase if too small
+                                                            // (assembler will crash if too small)
+  code_size2 = 35300 LP64_ONLY(+25000) WINDOWS_ONLY(+3200)  // simply increase if too small
+                                                            // (assembler will crash if too small)
```
tstuefe: Please remove all whitespace changes and reduce the patch to the absolutely necessary. The only changed line should be line 37.

tstuefe (approved):
This looks okay to me.

src/hotspot/cpu/x86/stubRoutines_x86.hpp
fthevenet: Done.

gnu-andrew: +1.  I think it's also worth a comment on the line above the `code_size2` definition to indicate why the `WINDOWS_ONLY` addition is there (e.g `// Windows ABI needs a little more space` as you've mentioned in the PR)

From reading the PR, it seems that the rare feature combination leads to this bug being hit, but the fact that this was also changed in implementing Virtual Threads suggests that is not the only possible cause.

In fact, you could just reuse the comments added by [8231349](https://github.com/openjdk/jdk/commit/3859faf183c241f124879d6a7264b43a6b42b418):
~~~
+  // AVX512 intrinsics add more code in 64-bit VM,
+  // Windows have more code to save/restore registers
~~~
(though feels like it should really be 'has more code')

gnu-andrew:
Please see my comments about adding a comment, now hidden under Thomas' comments.

Also I think we should wait until [JDK-8384158](https://bugs.openjdk.org/browse/JDK-8384158) is in and we can hopefully run this on Windows.  I've just approved it for 25u.

fthevenet:
> Please see my comments about adding a comment, now hidden under Thomas' comments.

Yes, it's a good idea.

 
> Also I think we should wait until [JDK-8384158](https://bugs.openjdk.org/browse/JDK-8384158) is in and we can hopefully run this on Windows. I've just approved it for 25u.

Sure thing.

gnu-andrew:
> > Also I think we should wait until [JDK-8384158](https://bugs.openjdk.org/browse/JDK-8384158) is in and we can hopefully run this on Windows. I've just approved it for 25u.
> 
> Sure thing.

If you make this PR target [pr/4451](https://github.com/openjdk/jdk17u-dev/pull/4451) instead of `master`, you should now be able to test on `windows-2022`.

gnu-andrew:
Sorry, I forgot to mention you also need to do a merge to actually bring in the workflow changes :-)

I just did this locally, where `jdk17u-dev` is my openjdk/jdk17u-dev remote and `fthevenet-17u-dev` is your fork.

~~~
$ git fetch -v jdk17u-dev
...
$ git merge jdk17u-dev/pr/4451
...
$ git diff fthevenet-17u-dev/JDK-8383659 -- .github/workflows/main.yml
diff --git a/.github/workflows/main.yml b/.github/workflows/main.yml
index 71bfb1cae45..0e60b8feb8b 100644
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -259,6 +259,7 @@ jobs:
     uses: ./.github/workflows/build-windows.yml
     with:
       platform: windows-x64
+      runs-on: windows-2022
       msvc-toolset-version: '14.44'
       msvc-toolset-architecture: 'x86.x64'
       configure-arguments: ${{ github.event.inputs.configure-arguments }}
@@ -271,6 +272,7 @@ jobs:
     uses: ./.github/workflows/build-windows.yml
     with:
       platform: windows-aarch64
+      runs-on: windows-2022
       msvc-toolset-version: '14.44'
       msvc-toolset-architecture: 'arm64'
       make-target: 'hotspot'
@@ -312,5 +314,4 @@ jobs:
     with:
       platform: windows-x64
       bootjdk-platform: windows-x64
-      runs-on: windows-2025
-
+      runs-on: windows-2022
~~~

Push the merge to your fork and we should hopefully get a clean Windows run as on https://github.com/openjdk/jdk17u-dev/pull/4451

gnu-andrew (approved):
Thanks for adding the comments. Code changes look good.

fthevenet:
> If you make this PR target [pr/4451](https://github.com/openjdk/jdk17u-dev/pull/4451) instead of `master`, you should now be able to test on `windows-2022`.

~I've changed the target for this PR to [pr/4451](https://github.com/openjdk/jdk17u-dev/pull/4451), but it doesn't seem to help. In hindsight, I believe this is normal since the tests are run on the original branch, prior to it being merged into the target. I think I need to rebase my PR onto a branch that include the change we want for it to be picked up in the GHA tests.
I'm thinking the cleanest way to do that is wait for pr/4451 to be merged, and then rebase this on master (and set the target back to `master`, of course)
Wdyt?~

EDIT: Looks like out our mutual comments collided :-)
I'll proceed as per your instructions above.

gnu-andrew:
> > If you make this PR target [pr/4451](https://github.com/openjdk/jdk17u-dev/pull/4451) instead of `master`, you should now be able to test on `windows-2022`.
> 
> ~I've changed the target for this PR to [pr/4451](https://github.com/openjdk/jdk17u-dev/pull/4451), but it doesn't seem to help. In hindsight, I believe this is normal since the tests are run on the original branch, prior to it being merged into the target. I think I need to rebase my PR onto a branch that include the change we want for it to be picked up in the GHA tests. I'm thinking the cleanest way to do that is wait for pr/4451 to be merged, and then rebase this on master (and set the target back to `master`, of course) Wdyt?~
> 
> EDIT: Looks like out our mutual comments collided :-) I'll proceed as per your instructions above.

Thanks. I don't know why I thought changing the target would magically merge the changes as well.  When the other PR gets merged, it will automatically flip the target back to `master`.

gnu-andrew:
Ah great, all builds now pass :)

fthevenet:
Yes, it all looks good now. 
Thanks for your help @gnu-andrew

GoeLin:
Hi [~fthevenet], 
thanks for fixing this issue.  
Could you please prefix the JBS bug title with [17u]?
Thanks!

gnu-andrew (approved):
Merged version still looks fine.

fthevenet:
Thanks @gnu-andrew

PR #4569, opened 2026-07-11
8341527: AVX-512 intrinsic for SHA3

missa-prime:
Hi all,

This pull request contains a backport of commit [9cfb0f7f](https://github.com/openjdk/jdk/commit/9cfb0f7f7ad31081c917be1eb0e39e2552e45382) from the [openjdk/jdk](https://git.openjdk.org/jdk) repository.

The commit being backported was authored by Ferenc Rakoczi on 29 Oct 2024 and was reviewed by Sandhya Viswanathan.

Thanks!

---------
- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

---------
### Progress
- \[ \] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- \[x\] Change must not contain extraneous whitespace
- \[x\] Commit message must refer to an issue
- \[ \] [JDK-8341527](https://bugs.openjdk.org/browse/JDK-8341527) needs maintainer approval

### Issue
 * [JDK-8341527](https://bugs.openjdk.org/browse/JDK-8341527): AVX-512 intrinsic for SHA3 (**Enhancement** - P3)

### Reviewers
 * [Sandhya Viswanathan](https://openjdk.org/census#sviswanathan) (@sviswa7 - no project role) 🔄 Re-review required (review applies to [a2e31e34](https://git.openjdk.org/jdk17u-dev/pull/4569/files/a2e31e34429ee03696c3dc0b0ce2c4000278da53))

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/4569/head:pull/4569` \
`$ git checkout pull/4569`

Update a local copy of the PR: \
`$ git checkout pull/4569` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/4569/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 4569`

View PR using the GUI difftool: \
`$ git pr show -t 4569`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk17u-dev/pull/4569.diff">https://git.openjdk.org/jdk17u-dev/pull/4569.diff</a>

</details>
<details><summary>Using Webrev</summary>

[...]

PR #4603, opened 2026-08-08
8341527: AVX-512 intrinsic for SHA3

GoeLin:
This backport is in large parts of its edits based on missa-prime's PR #4569.
Also his previous backports are important for this.
I based this backport on the change in 21.

This is a try to stay closer to the sources in 21, though,
to make later backports that are planned already,
easier.

The second commit contains the resolved chunks as well 
as adaptions to different stubGenerator in 17:

src/hotspot/cpu/x86/assembler_x86.cpp
src/hotspot/cpu/x86/assembler_x86.hpp
Resolved due to context.
src/hotspot/cpu/x86/macroAssembler_x86.hpp
Resolved due to context.
Omitted 'using' statements. Methods not in 17.
src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
Resolved due to context.

src/hotspot/cpu/x86/stubGenerator_x86_64.hpp
This file does not exist in 17. As the stubGenerator is not 
in a header, it is not easy to split the implementation
into several files. But this is done for this change in 21,
as well as later ML-KEM and ML-DSA intrinsics. 

I propose to just in include stubGenerator_x86_64_sha3.cpp
in stubGenerator_x86_64.cpp. This compiles with very few 
changes, amon other a guard to avoid that 
stubGenerator_x86_64_sha3.cpp is compiled stand-alone.

The third commit has additional coding taken from 
missa-prime's PR #4569.  I tried to place the new
instructions in assembler_x86.c|hpp at the same locations
as in the current 21 file.

Update:
I tweaked not compiling the new file: This can be specified in the makefiles.
The other approach made problems on windows.

---------
- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

---------
### Progress
- \[x\] Change must not contain extraneous whitespace
- \[x\] Commit message must refer to an issue
- \[x\] [JDK-8341527](https://bugs.openjdk.org/browse/JDK-8341527) needs maintainer approval
- \[ \] Change must be properly reviewed (2 reviews required, with at least 2 [Reviewers](https://openjdk.org/bylaws#reviewer))

### Issue
 * [JDK-8341527](https://bugs.openjdk.org/browse/JDK-8341527): AVX-512 intrinsic for SHA3 (**Enhancement** - P3 - Approved)

### Reviewers
 * [David Briemann](https://openjdk.org/census#dbriemann) (@dbriemann - Committer) 🔄 Re-review required (review applies to [50e00acd](https://git.openjdk.org/jdk17u-dev/pull/4603/files/50e00acd91c6ed148dfb90e47ca4d876b5db3cf0))
 * [Martin Doerr](https://openjdk.org/census#mdoerr) (@TheRealMDoerr - **Reviewer**)

### Contributors
 * Mohamed Issa `<missa@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk17u-dev.git pull/4603/head:pull/4603` \
`$ git checkout pull/4603`

`$ git checkout pull/4603` \
`$ git pull https://git.openjdk.org/jdk17u-dev.git pull/4603/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 4603`

`$ git pr show -t 4603`

</details>

<a href="https://git.openjdk.org/jdk17u-dev/pull/4603.diff">https://git.openjdk.org/jdk17u-dev/pull/4603.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk17u-dev/pull/4603#issuecomment-5290645743)
</details>


src/hotspot/cpu/x86/macroAssembler_x86.hpp
```diff
@@ -1376,6 +1379,9 @@ class MacroAssembler: public Assembler {
   void vpsraw(XMMRegister dst, XMMRegister nds, XMMRegister shift, int vector_len);
   void vpsraw(XMMRegister dst, XMMRegister nds, int shift, int vector_len);
 
+  void evpsrad(XMMRegister dst, XMMRegister nds, XMMRegister shift, int vector_len);
+  void evpsrad(XMMRegister dst, XMMRegister nds, int         shift, int vector_len);
+
```
dbriemann: These are declared but not defined. I don't think they are needed for the PR. Did this slip in?

src/hotspot/cpu/x86/assembler_x86.cpp
```diff
@@ -11080,3 +11153,28 @@ void InstructionAttr::set_address_attributes(int tuple_type, int input_size_in_b
...
+}
+
+void Assembler::evpermt2q(XMMRegister dst, XMMRegister nds, XMMRegister src, int vector_len) {
+  assert(VM_Version::supports_evex() && (vector_len == Assembler::AVX_512bit || VM_Version::supports_avx512vl()), "");
+  InstructionAttr attributes(vector_len, /* vex_w */ true, /* legacy_mode */ false, /* no_mask_reg */ true, /* uses_vl */ true);
+  attributes.set_is_evex_instruction();
+  int encode = vex_prefix_and_encode(dst->encoding(), nds->encoding(), src->encoding(), VEX_SIMD_66, VEX_OPCODE_0F_38, &attributes);
+  emit_int16(0x7E, (0xC0 | encode));
+}
+
```
dbriemann: `git diff --check` found a new blank line at EOF.

GoeLin:
Hi @dbriemann 
thanks for having a look. I pushed a change implementing your requests, and also merged head.

dbriemann (approved):
LGTM 👍

make/hotspot/lib/CompileJvm.gmk
```diff
@@ -125,6 +125,7 @@ endif
 ifeq ($(call isTargetCpu, x86), true)
   JVM_EXCLUDE_PATTERNS += x86_64
 else ifeq ($(call isTargetCpu, x86_64), true)
+  JVM_EXCLUDE_FILES += stubGenerator_x86_64_sha3.cpp
```
TheRealMDoerr, line 128: Please explain. Why do we add a file which we don't want to compile?

src/hotspot/cpu/x86/assembler_x86.cpp
```diff
@@ -7486,6 +7520,34 @@ void Assembler::evpxorq(XMMRegister dst, XMMRegister nds, Address src, int vecto
...
+  attributes.set_is_evex_instruction();
+  attributes.set_embedded_opmask_register_specifier(mask);
+  if (merge) {
+    attributes.reset_is_clear_context();
+  }
+  vex_prefix(src, nds->encoding(), dst->encoding(), VEX_SIMD_66, VEX_OPCODE_0F, &attributes);
+  emit_int8((unsigned char)0xEF);
+  emit_operand(dst, src);
+}
+
```
TheRealMDoerr, line 7530: Upstream version has the different evpxorq functions in a different order:
```
void Assembler::evpxorq(XMMRegister dst, KRegister mask, XMMRegister nds, XMMRegister src, bool merge, int vector_len) {
void Assembler::evpxorq(XMMRegister dst, KRegister mask, XMMRegister nds, Address src, bool merge, int vector_len) {
void Assembler::evpxorq(XMMRegister dst, XMMRegister nds, XMMRegister src, int vector_len) {
void Assembler::evpxorq(XMMRegister dst, XMMRegister nds, Address src, int vector_len) {
```
Your version has swapped the first 2 with the second 2.

TheRealMDoerr (commented):
Looks basically good. Thanks for backporting!

make/hotspot/lib/CompileJvm.gmk
dbriemann: Let me answer, because I had to figure that out too. I think this prevents the build from creating an object file for `stubGenerator_x86_64_sha3.cpp`. However the file is directly included at the end of `stubGenerator_x86_64.cpp` and implicitly built into that object file.

TheRealMDoerr: Not a nice solution, but I could live with it. I have no idea what the other approach was and why it caused problems on Windows.

GoeLin:
Hi @dbriemann, @TheRealMDoerr 
yes, you are right, I include the one file in the other.
I tried to explain this in the description of the PR:

21 has refactored the stubs of x86 to have a header file that declares class StubGenerator.
The implementation of the stubs are in lots of different files stubGenerator_<xyz>.cpp
that include the header with the class definition.

17 declares class StubGenerator in a .cpp file.
Stubs are implemented in macroAssembler_<xyz>.cpp files.  
There is no natural way to make stubGenerator_x86_64_sha3.cpp added by this change 
to know about class StubGenerator.
In #4569, missa-prime proposed to transform stubGenerator_x86_64_sha3.cpp
into macroAssembler_x86_64_sha3.cpp and streamline
the code with jdk17.  I tried to review that, but you can not compare the 
two files with tooling. E.g. all lines differ as "__" has been removed.
I think this setup will unnecessarily complicate backports to macroAssembler_x86_64_sha3.cpp.
You must rely on your eyes to compare assembler coding, this is error prone.
Therefore I propose the setup with #including stubGenerator_x86_64_sha3.cpp
for this and the follow-up changes.

> I have no idea what the other approach was

I first tried to avoid stand alone compilation of stubGenerator_x86_64_sha3.cpp
by preprocessor guards, but that failed as windows has a check for precompiled.hpp.
So I use make to avoid compilation, that's much better anyways.
See also commit [better tweak to exclude new file from compiling](https://github.com/openjdk/jdk17u-dev/pull/4603/changes/0f22298d3519402a65cdde2472ac6670887e8347)

Best, Goetz.

src/hotspot/cpu/x86/assembler_x86.cpp
GoeLin: Fixed this.  Now looks similar to https://github.com/openjdk/jdk21u-dev/commit/4598c35c2f6cd2b440395c19e262affa922a0a8b

Thanks for catching this!

TheRealMDoerr (approved):
Ok, thanks for the explanation! If your current solution is easier to maintain, I'll be ok with it.

jerboaa:
This is a pretty hairy change and would like for an additional reviewer to have a look. @rwestrel @adinn @missa-prime Could you take a look, please?

jerboaa:
Has anyone done the performance tests on SHA3 on AVX512 on JDK 17 with this?

adinn:
> This is a pretty hairy change and would like for an additional reviewer to have a look. @rwestrel @adinn @missa-prime Could you take a look, please?

I just read the comment about using `#include`  to include a `.cpp` file (`stubGenerator_x86_64_sha3.cpp`) and thought yeeurch. That sounds like a horrible solution. I'll take a look.

src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
```diff
@@ -7811,3 +7817,7 @@ void StubGenerator_generate(CodeBuffer* code, bool all) {
   }
   StubGenerator g(code, all);
 }
+
+#undef __
+
+#include "stubGenerator_x86_64_sha3.cpp"
```
adinn, line 7823: I think this is a terrible way to achieve the desired result of making the class declaration visible to the method implementations. This will only get worse as we backport more of the PQC code.

It would probably be better to start by backporting JDK-8293188, which split the class definition out into a separate header, before backporting any of the PQC code. yes, it's a small amount of extra, initial churn but it will make the rest of the backporting cleaner and will leave us with a file organization that makes sense. Please at least try this alternative first to see how it looks.

GoeLin: Hi,
what I intend is to be able to compare the ssh3 assembly code in 17 to that in 21.  For this I would like to keep the new file as-is.  Jeah, JDK-8293188 might do the job in a better way.  I'll look into this at a later point, though.  Won't make the October update anyways any more I guess.
Thanks for looking into this!

adinn:
> What I intend is to be able to compare the ssh3 assembly code in 17 to that in 21. For this I would like to keep the new file as-is. Yeah, JDK-8293188 might do the job in a better way. I'll look into this at a later point, though. Won't make the October update anyways any more I guess.

Well, thank you for trying my suggested alternative. I'd really prefer to avoid this version going in if at all possible.

I'm not quite clear why you bring up the question of any code comaprison. I don't think splitting out the class into a header will require you to make significant changes to `stubGenerator_x86_64_sha3.cpp`. Indeed, it should mean less changes to that file than the current approach. Also, I would not expect either approach to result in different generated code.

GoeLin:
Hi @adinn, 
the natural way of implementing this in 17 would be to move this into a macroAssember_x86_64_sha3 file, as proposed by missa in #4569, see also the first line in the description of this PR.  I did not like that solution when I tried to review the assembler coding.   So my goal of this implementation was to keep the stubGenerator file so that the assembly looks exactly as in 21.

GoeLin:
I'll eventually pick this up later again.

